PR #805, opened 2023-10-13
feat(devnet): run pg-queue migrations on startup

hussein-aitlahcen:


benluelo:
closing as no longer relevant, voyager now runs migrations itself on startup

PR #1251, opened 2024-02-01
feat: begin app development

o-az:
- [x] app/ dir in monorepo root
- [x] contains a SvelteKit app
- [x] The SvelteKit app should use adapter-static, no server-side components.
- [x] Create an app package in Nix such that we can do nix build .#app.
- [x] Preview/prod deployments in the same way as site/
- [x] Configure Cloudflare such that app.union.build points to the prod deployment
- [x] Include Tailwind in the setup

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -98,9 +98,26 @@
       flake = false;
     };
   };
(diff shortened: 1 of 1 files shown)


U9tica:
UNION MONOREPO IS GREAT

PR #1773, opened 2024-04-17
DO NOT MERGE: Ethereum light client track justified/finalized header

### LINKED ISSUE #1108: Implement `justified|finalized` header tracking in ethereum lc
Ethereum light client should be capable of tracking both of the headers and it should be configurable

### LINKED ISSUE #1127: Create 2 different lodestar options for devnet
We need two lodestar running options, one for justified and one for finalized

aeryz:
Note that right now, `voyager` is configured to select justified/finalized at compile time via feature gating. This is no good though. It's being done as an easy/quick solution. We want to add a custom spec parameter called `CHECKPOINT_ROOT_INDEX` and return it via `/eth/v1/config/spec` RPC. We can easily do it by modifying `lodestar` but before doing that I want to be sure that there is no easy way of doing it (via a cli parameter or config file). I created [this discussion](https://github.com/ChainSafe/lodestar/discussions/6683) regarding this,

**EDIT:** The discussion is closed and unfortunately we need to add it to the spec in the code. I will do it in another PR though. See #1784 

Closes #1108 
Closes #1127


aeryz:
Intent-based bridging will make this unnecessary

PR #2034, opened 2024-06-04
feat(voyager): start adding prometheus metrics

benluelo:
adds basic prometheus metrics to voyager. currently the only metric is the queue msg processing time. more will be added soon.

`lazy_cell` has been stabilized, we just need to bump our rust version. i'll do that soon(ish).


benluelo:
superceded by #4548

PR #2793, opened 2024-08-20
feat(scroll): upgrade verifier and light client to v3

### LINKED ISSUE #1830: Investigate storing the l1 state root in the scroll light client
It may be a good idea to directly store the l1 state root in the scroll light client, to make it easier to verify against (say via an L3 that settles on scroll). This would be duplication though, since it would also be possible to query the l1 client directly - we should investigate if this is worth the added complexity in the scroll light client, or if the complexity should be offloaded to whatever is verifying against this l1 state root.

### LINKED ISSUE #2791: Scroll - Update to V3
We currently are on V2 of scroll batch headers. The V3 header includes the latest block timestamp, which allows us to ditch most of the code that was used to reconstruct the chunks to extract the latest block timestamp. See https://scrollzkp.notion.site/Darwin-Upgrade-4-Note-for-Partners-e96f490b211e4dbda9f4f523242ef34b

### LINKED ISSUE #2792: Voyager - Scroll V3
Following #2791

hussein-aitlahcen:
fixes #2791 
fixes #2792 
fixes #1830
towards #2792


benluelo (approved):

benluelo (approved):
🚀

PR #2823, opened 2024-08-23
fix(ci): use check-requsted event trigger

PoisonPhang:


benluelo (approved):

PoisonPhang:
Workflow is no longer required

PR #2846, opened 2024-08-27
feat: multicall token balances query

o-az:


app/src/lib/assets/base64.txt
o-az, line 1: this file was never meant to be in `/static` where I originally placed it. While testing the multicall function in the cli `bun src/lib/queries/balance/evm/multicall.ts` it blocked me from making the cli call so I'm fixing its location issue in the same pr might as well.

app/package.json
o-az, line 1: encountered weird type errors and npm security warnings which were resolved by updating wagmi and viem deps

cor (approved):

cor (approved):

PR #2861, opened 2024-08-29
feat(app): packets icon

### LINKED ISSUE #2844: Packets icon for packets in explorer sidebar
image

Currently, the packets icon is the same as the transfers icon. Tucker designed one specifically for packets though, we should swap it out

Swepool:
closes #2844 

Now u can use as svelte component

```js
import UnionIcon from "$lib/components/union-icons/index.svelte"

<UnionIcon theme="mono" name="transfers" class="mr-2 size-5" />
```
or import  and use in TS/JS

```js
import { UnionIcons } from "$lib/components/union-icons/union-icons.ts"

{ route: "packets", icon: UnionIcons.packet.variants.color, description: "All packets" },
```


cor (changes requested):
Looks nice, but does this restructure tree-shake well? @o-az

PR #2867, opened 2024-08-29
feat(app):  make recipient editable with reset

### LINKED ISSUE #2770: Make receiver editable on transfer page
Selecting a destination chain should not require a wallet to be connected for that destination. The recipient logic should be identical to the logic on the faucet page:

Image

If a wallet is connected, we prefill it with the address.

Image

If the user edits it, a "reset" button appears. In this case, I recommend making the label _"set to connected wallet"_ rather than just "reset".

As part of this PR, I also recommend changing all instances of "recipient" to "reciever" as that matches all other naming of receiver.

Swepool:
closes #2770


app/src/routes/transfer/(components)/transfer-form.svelte
```diff
@@ -146,9 +147,9 @@ let recipient = derived([toChain, userAddr], ([$toChain, $userAddr]) => {
 })
```
cor, line 147: above here, rename `recipient` to `connectedWalletReceiver`

cor (changes requested):
Some comments, in addition: selecting a destination chain should no longer require a wallet to be connected, as you can fill in any address for the destination chain. Only require wallets to be connected for the source chain. Because of this, it's important that we do full verification of validity of the destination address

cor (changes requested):
Doesn't work, transferring from stride -> arbitrum with a manual receiver put in yields the following incorrect transfer:

https://app.union.build/explorer/transfers/015096DB79FF0D2D019306E2CF0856A622C346765E250E89B42BB1C227080DF0

whereas it does work when transferring from the current staging UI:

https://staging.app.union.build/explorer/transfers/F61C997FD42FAC9B68A9F54BC2FC5041286B77E49029F5781AE429526292CDE1

Perhaps there is an issue with the way in which you normalize the custom receiver address?

PR #2885, opened 2024-09-01
feat(docs): graphql demo

o-az:


cor:
It's nice but ideally we have have something that just parses all ```gql blocks and gives them a play button

cor:
this is no longer relevant as we have graphiql integrated

PR #2889, opened 2024-09-02
movement improvements

aeryz:


move/move-ibc/sources/ibc.move
```diff
@@ -1230,120 +928,481 @@ module IBC::Core {
...
+        )
+    }
+
+    public fun set_supported_versions(
+        supported_versions: vector<connection_end::Version>,
+        dst: &mut vector<connection_end::Version>
+    ) {
+        assert!(vector::length(dst) == 0, E_VERSION_MUST_BE_UNSET);
+        vector::append(dst, supported_versions);
+    }
```
benluelo, line 988: is there a reason not to just return a new vector here?

move/move-ibc/sources/ibc.move
```diff
@@ -1230,120 +928,481 @@ module IBC::Core {
...
+    public fun is_supported_version(
+        supported_versions: &vector<connection_end::Version>,
+        version: &connection_end::Version
+    ): bool {
+        let (supported_version, found) = find_supported_version(supported_versions, version);
+        if(found && verify_proposed_version(&supported_version, version)) {
+            return true
+        };
+        false
+        // found && verify_proposed_version(&supported_version, version)
```
benluelo, line 999: ?

move/move-ibc/sources/ibc.move
```diff
@@ -1230,120 +928,481 @@ module IBC::Core {
...
+            // if (byte < 0x61 || byte > 0x7A) { // ASCII values for 'a' and 'z'
+            //     return false;
+            // };
+            if (byte > 0x40 && byte < 0x5b) {
+                return false
+            };
+            i = i + 1;
+        };
+        true
+    }
```
benluelo, line 1356: the reason we did this in the solidity contract is that we re-encode the port (contract address) as hex when indexing into the capabilities table, and if you create the channel with a checksummed address as the port id then it doesn't match and you can't actually send any packets on the channel. i'm not sure how this will work in move, but do keep it in mind

move/move-ibc/sources/ibc.move
```diff
@@ -1230,120 +928,481 @@ module IBC::Core {
...
-        claim_capability(
-            IBCCommitment::channel_capability_path(port_id, channel_id),
-            signer::address_of(ibc_app),
+
+    public fun verify_channel_state(
+        connection: &ConnectionEnd,
+        height: height::Height,
+        proof: vector<u8>,
+        port_id: String,
+        channel_id: String,
```
benluelo, line 1377: does move have any concept of "zero cost newtypes"? might be nice to use something like that for the various id types instead of string directly.

PR #2904, opened 2024-09-03
feat: extract docs into own static site

o-az:


cor (changes requested):
image

Looks good, but the assets require some more thought. the assets dir in docs contains many assets like headshots that are only used on the main site

cor (approved):

PR #2944, opened 2024-09-12
Hubble reorg handling

qlp:


hubble/src/indexer/api.rs
```diff
@@ -0,0 +1,69 @@
...
+use sqlx::Postgres;
+use time::OffsetDateTime;
+use tokio::{sync::mpsc::Sender, task::JoinSet};
+
+#[derive(Debug, thiserror::Error)]
+pub enum FetcherError {
+    #[error("error fetching message: {0}")]
+    ErrorFetchingMessage(Report),
+    #[error("received unexpected height: {0} (expecting {1})")]
+    RecievedUnexpectedHeight(BlockHeight, BlockHeight),
```
KaiserKarel, line 14:
```suggestion
    ReceivedUnexpectedHeight(BlockHeight, BlockHeight),
```

hubble/src/indexer/fetcher.rs
```diff
@@ -0,0 +1,212 @@
...
+            FetchMessage::Fetched(block_handle) => {
+                debug!("fetched: {}", block_handle.height()?);
+
+                if expected_next_height != block_handle.height()? {
+                    error!(
+                        "unexpected: {} (expecting {})",
+                        block_handle.height()?,
+                        expected_next_height
+                    );
+                    return Err(FetcherError::RecievedUnexpectedHeight(
```
KaiserKarel, line 111:
```suggestion
                    return Err(FetcherError::ReceivedUnexpectedHeight(
```

hubble/src/indexer/fetcher.rs
```diff
@@ -0,0 +1,212 @@
...
+                Some(block_handle) => {
+                    debug!("{}: handling", block_handle.height()?);
+
+                    if next_height != block_handle.height()? {
+                        error!(
+                            "unexpected: {} (expecting {})",
+                            block_handle.height()?,
+                            next_height
+                        );
+                        return Err(FetcherError::RecievedUnexpectedHeight(
```
KaiserKarel, line 159:
```suggestion
                        return Err(FetcherError::ReceivedUnexpectedHeight(
```

hubble/src/indexer/fetcher.rs
```diff
@@ -0,0 +1,212 @@
...
+
+    async fn run_to_tip(&self, fetcher_client: &T) -> Result<(), Report> {
+        loop {
+            let next_height = self.next_height().await?;
+            info!("{}: fetching", next_height);
+
+            match fetcher_client
+                .fetch(BlockReference::Height(next_height))
+                .await?
+            {
```
KaiserKarel, line 149: So nit, this now does it sequentially again (we're no longer fetching in batches and inserting as fast as they come in). That actually means syncing from scratch takes a very long time (note some chains produces blocks sub 1 second, so the amount of blocks is quite big and keeps growing very quickly).

hubble/src/indexer/dummy/config.rs
```diff
@@ -0,0 +1,24 @@
...
+use crate::indexer::{api::IndexerId, Indexer};
+
+#[derive(Clone, Debug, serde::Deserialize)]
+pub struct Config {
+    pub indexer_id: IndexerId,
+    pub start_height: BlockHeight,
+}
+
+impl Config {
+    pub async fn build(self, pg_pool: PgPool) -> Result<Indexer<DummyFetcherClient>, Report> {
```
KaiserKarel, line 15: Does this need to be async?

hubble/src/indexer/fetcher.rs
```diff
@@ -0,0 +1,212 @@
...
+                            .await?;
+
+                        let mut expected_next_height = slice.start;
+
+                        while let Some(message) = rx.recv().await {
+                            expected_next_height = self
+                                .handle_fetch_message(message, expected_next_height)
+                                .instrument(info_span!("store"))
+                                .await?
+                        }
```
KaiserKarel, line 88: More of a style issue, but it is better to have fetch_range return a Stream vs using a channel to pass messages. That will allow you to handle messages concurrently, as well as make debugging easier (slap a .inspect on the stream).

API would then be:

```
let stream = last_finalized.fetch_range(...).instrument(...);
stream.for_each(self.handle_fetch_message).await;
```

hubble/src/indexer/fetcher.rs
qlp: we still have this approach:
- first: run to finalized
- then: run to tip

hubble/src/indexer/dummy/config.rs
qlp: want to have it consistent with other indexers (where we do some async initialization). 

ultimately (once all indexers are migrated) I'l remove the async from all indexers.

hubble/src/indexer/api.rs
```diff
@@ -0,0 +1,149 @@
...
+
+impl BlockRangeUtil for BlockRange {
+    fn to_interval(&self) -> BlockInterval {
+        BlockInterval(self.clone())
+    }
+
+    fn len(&self) -> u64 {
+        self.end - self.start
+    }
+}
```
benluelo, line 69: any reason this is a trait instead of implemeted directly on the BlockInterval type?

hubble/src/indexer/eth/context.rs
```diff
@@ -0,0 +1,23 @@
...
+            "urls: {}",
+            self.urls
+                .iter()
+                .enumerate()
+                .map(|(index, url)| format!("{}: {}", index, url.as_str().to_string()))
+                .collect::<Vec<_>>()
+                .join(", ")
+        )
+    }
+}
```
benluelo, line 23: nit: what are these display impls used for? seems like this should actually be debug?

qlp: formatting log messages. yes, could also be debug.

can I assume: if only used in logging, then only implement Debug?

hubble/src/indexer/api.rs
qlp: no, leftover before introducing BlockInterval. I'll fix

hubble/src/indexer/fetcher.rs
qlp: done

benluelo (approved):

PR #2981, opened 2024-09-19
fix(hubble): bug missing logging spans

qlp:


hubble/src/indexer/mod.rs
```diff
@@ -56,25 +56,34 @@ where
...
                     let fetcher_client_clone = fetcher_client.clone();
-                    join_set.spawn(async move { self_clone.run_fixer(fetcher_client_clone).await });
+                    join_set.spawn(
+                        async move { self_clone.run_fixer(fetcher_client_clone).await }
+                            .instrument(info_span!("fixer")),
+                    );
 
                     if let EndOfRunResult::Exit = self
                         .handle_end_of_run(&mut join_set, fetcher_client)
+                        .instrument(info_span!("exiter"))
```
benluelo, line 86: probably good to include some context in these spans as well? something like `info_span!(context = %self.context, "fetcher")`

benluelo: this behaviour is not guaranteed, and may change depending on how the tasks are spread across threads. i believe you can use [`in_current_span`](https://docs.rs/tracing/latest/tracing/trait.Instrument.html#method.in_current_span) to guarantee this though

benluelo: ah interesting, looks solid then 👍

benluelo: i think the part i was thinking about was:

> However, if "my_future" is disabled, the spawned task will *not* be in any span.

where if you want the future to *always* be in the current span, you need to use `in_current_span`

benluelo (approved):

PR #2986, opened 2024-09-20
feat: sharable graphql queries & doc theme

o-az:


docs/docs.nix
```diff
@@ -79,6 +79,37 @@
...
+                ${ensureAtRepositoryRoot}
+                
+                rm -rf docs/generated/rust
+                mkdir -p docs/generated/rust
+
+                ${lib.concatMapStrings (pkg: ''
+                  cargo rustdoc \
+                    --release \
+                    --all-features \
+                    --package='${pkg}' \
```
benluelo, line 103: you can pass -p more than once at a time. use https://doc.rust-lang.org/nightly/rustdoc/unstable-features.html#--enable-index-page-generate-a-default-index-page-for-docs as well for a root level directory with all packages.

docs/docs.nix
```diff
@@ -79,6 +79,37 @@
             '';
           };
         };
+        generate-rust-docs = {
```
benluelo, line 82: this needs to be done in a derivation, such that it will be run in CI and ensure that docs are always up to date: https://crane.dev/API.html#cranelibcargodoc

cor (approved):

PR #2988, opened 2024-09-20
chore: improve docs across the workspace & small fix

benluelo:
- move to `#[model]` for some types in `voyager-message`
- small fix for handling invalid plugin messages as fatal
- improve docs in `voyager-message` and all dependencies
- clean up in `voyager-message`, `chain-utils`, and `unionlabs` (remove a lot of unused code)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -107,6 +107,9 @@ repository   = "https://github.com/unionlabs/union"
 disallowed_types = "deny"
 module_inception = "allow" # it's useful sometimes
 
+[workspace.lints.rustdoc]
+broken-intra-doc-links = "deny"
+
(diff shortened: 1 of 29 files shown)


Cargo.toml
```diff
@@ -107,6 +107,9 @@ repository   = "https://github.com/unionlabs/union"
 disallowed_types = "deny"
 module_inception = "allow" # it's useful sometimes
 
+[workspace.lints.rustdoc]
+broken-intra-doc-links = "deny"
```
benluelo, line 111: this doesn't fully work, i need to look into the semantics of this further

PoisonPhang (approved):

PR #2999, opened 2024-09-21
feat(voyager): split out voyager-core

benluelo:


lib/voyager-message/Cargo.toml
```diff
@@ -49,6 +49,7 @@ tracing                        = { workspace = true }
 tracing-subscriber             = { workspace = true, features = ["json"] }
 typenum                        = { workspace = true }
 unionlabs                      = { workspace = true, features = ["ethabi"] }
+voyager-core.workspace         = true
```
aeryz, line 52: I think we should go with `{ workspace = true }` everywhere

aeryz (dismissed):

lib/voyager-message/Cargo.toml
benluelo: yes i agree, however `cargo add` defaults to .workspace = true :(

benluelo: fixed all existing occurences of .workspace; ill look into formatting solutions for this

benluelo: https://github.com/unionlabs/union/issues/3007

lib/voyager-message/src/rpc/server.rs
```diff
@@ -60,8 +57,8 @@ pub struct ServerInner {
 // }
 
 // impl IbcStateCacheKey<'static> {
-//     pub fn new_owned(chain_id: ChainId<'static>, path: Path, height: Height) -> Self {
-//         Self {
+//     pub fn new_owned(chain_id: ChainId<'static>, path: Path, height: Height)
+// -> Self {         Self {
 //             chain_id,
 //             path: Cow::Owned(path),
 //             height,
```
hussein-aitlahcen, line 64: do we keep all this comments?

hussein-aitlahcen (approved):

lib/voyager-message/src/rpc/server.rs
benluelo: i was keeping it for now because i'll likely use this for more manageable cache keys soon (the only changes here is comment wrapping)

PR #3025, opened 2024-09-26
feat(hubble): implement token import

PablitoAmaral:


hubble/src/token_list.rs
```diff
@@ -0,0 +1,152 @@
...
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct TokenInsert {
+    pub chain_id: i64,
+    pub denom: String,
+    pub display_symbol: String,
+    pub decimals: i64,
+    pub logo_uri: Option<String>,
+    pub display_name: String,
+    pub gas_token: bool, // always false?
```
qlp, line 137: we should insert `false` on new records and keep the existing value if we're updating. so I think you best remove the field here and update the insert statement

hubble/src/postgres.rs
```diff
@@ -605,3 +605,80 @@ pub async fn insert_client_mapping<'a, A: Acquire<'a, Database = Postgres>>(
...
+    Ok(chain_ids_and_ids)
+}
+
+pub async fn insert_or_update_tokens<'a, A: Acquire<'a, Database = Postgres>>(
+    db: A,
+    tokens: &[(i64, String, String, i64, Option<String>, String, bool)],
+) -> sqlx::Result<()> {
+    let mut conn = db.acquire().await?;
+
+    let (chain_ids, denoms, display_symbols, decimals, logo_uris, display_names, gas_tokens): (
```
qlp, line 630: personally I wouldn't do the batch insert, but insert one-by-one: keeps come simpler and in this (low-volume) scenario good enough.

... but feel free to keep it as is

hubble/src/postgres.rs
```diff
@@ -605,3 +605,80 @@ pub async fn insert_client_mapping<'a, A: Acquire<'a, Database = Postgres>>(
...
+pub async fn get_chain_ids_and_ids<'a, A: Acquire<'a, Database = Postgres>>(
+    db: A,
+) -> sqlx::Result<Vec<(String, i32)>> {
+    let mut conn = db.acquire().await?;
+
+    let rows = sqlx::query!("SELECT chain_id, id FROM v0.chains")
+        .fetch_all(&mut *conn)
+        .await?;
+
+    let chain_ids_and_ids: Vec<(String, i32)> =
```
qlp, line 618: try to use a HashMap<String, i32> here. will make the code using it easier

qlp (commented):
looking good: some small suggestions.

general note: add some info and debug logging. normally the app runs with just info logging and we enable debug when there are issues. think what you would like to see: 

with info you want to get confirmation that the process is running as expected
with debug you want to see key decision points in the process

qlp (approved):

PR #3026, opened 2024-09-26
feat: rewrite queue-msg normalization pass

### LINKED ISSUE #3003: optimize queue-msg normalization pass
the normalization pass for `queue-msg` is quite inefficient. since this function is called on every message pushed into the queue, *and* on the output of every optimization pass, this needs to be optimized. it was copied from the previous implementation which had each step as a separate pass, but it can now be consolidated into one:

https://github.com/unionlabs/union/blob/86265eb82991586ec571b2540613589216c48a13/lib/queue-msg/src/normalize.rs

the main two inefficiencies are:

1. allocates new `Vec`s for every step for both the parent id indexes and output `Op`s
2. descends into the `Op`s multiple times

fixing these will essentially require a bottom-up rewrite of the normalization pass.

requirements:

- all existing (non-commented-out) tests must still pass
  - if a semantic bug in one of these tests is found, then keep that test as a failure case and write a new test for the correct semantics.
- benchmarks for comparison with current and new

> [!NOTE]
> i've marked this as E-easy since this is a very isolated component. i am willing to provide mentoring on this task if desired.

benluelo:
closes #3003

results:

```
normalize               time:   [15.757 µs 15.946 µs 16.192 µs] // old
normalize2              time:   [5.5628 µs 5.5822 µs 5.6047 µs] // new
```

also removed some unused `Op` variants.


lib/queue-msg/src/lib.rs
```diff
@@ -408,16 +376,16 @@ impl<T: QueueMessage> Op<T> {
...
+                // Op::Repeat { times: None, msg } => {
+                //     Ok(Some(seq([*msg.clone(), repeat(None, *msg)])))
+                // }
+                // Op::Repeat {
+                //     times: Some(times),
+                //     msg,
+                // } => Ok(Some(seq([*msg.clone()].into_iter().chain(
+                //     // if times - 1 > 0, queue repeat with times - 1
+                //     NonZeroU64::new(times.get() - 1_u64).map(|times| repeat(Some(times), *msg)),
+                // )))),
```
hussein-aitlahcen, line 388: we ditch `repeat` and `race` completely?

benluelo: yes, forgot to remove this part fully - but imo these can both be implemented more efficiently for whatever specific usecase the functionality is needed for. removing these also removes significant complexity from the `Op` type and reduces potential nesting.

hussein-aitlahcen (approved):

hussein-aitlahcen (approved):

PR #3034, opened 2024-09-27
fix(ci): use nix_conf to load caches rather than nix develop

PoisonPhang:
Speedy deployment example
https://github.com/unionlabs/union/actions/runs/11072802396/job/30768114276


benluelo (approved):

o-az (changes requested):
rm the other `nix develop` occurances?

o-az (approved):

PR #3078, opened 2024-10-08
chore(move-ibc): docs

aeryz:


move/move-ibc/README.md
benluelo, line 1: Looks good, but the code samples will bitrot very quickly - perhaps we can host the generated move docs and link to that?

aeryz: I still want to provide a walkthrough because even if we have link to the generated docs, people will still search for a walkthrough. And since these are core IBC functions, I don't think they will frequently change.

benluelo: We'll just need to be diligent then 👍

benluelo (approved):

PR #3095, opened 2024-10-11
fix(ceremony): temp use blue

Swepool:


hussein-aitlahcen (approved):

cor (approved):
for clarification, this is for `app`, not for `ceremony` (despite PR title)

PR #3106, opened 2024-10-15
Aptos Move IBC misbehaviours and `movefmt`

aeryz:
- Introduced `submit_misbehaviour` entrypoint.
- Added misbehaviour check during `update_client`.
- Added `movefmt` to `nix fmt`.
- Formatted all move files.


Caglankaan (approved):
zlgtm!

PR #3115, opened 2024-10-17
fix(direnv): skip lfs smudge

cor:


.envrc
```diff
@@ -1,3 +1,4 @@
 [[ -f .envrc.local ]] && source_env .envrc.local
 use flake
-export DIRENV_WARN_TIMEOUT=1m
+export DIRENV_WARN_TIMEOUT=2m
+export GIT_LFS_SKIP_SMUDGE=true
```
benluelo, line 4: leave a comment explaining why perhaps

Issue #2997, opened 2024-09-21
organize light client implementations


benluelo:
light client implementations are currently scattered across the repo. these should be consolidated; i suggest `lightclients/{client-type}/{ibc-interface}`, as it matches voyager:

```
lightclients/
|-- arbitrum
|   `-- ibc-go-v8-08-wasm
|-- berachain
|   `-- ibc-go-v8-08-wasm
|-- cometbls
|   |-- ibc-solidity
|   |-- aptos-move
|   |-- ibc-go-v8-native
|   `-- ibc-go-v8-08-wasm
|-- ethereum
|   `-- ibc-go-v8-08-wasm
|-- evm-in-cosmos
|   `-- ibc-go-v8-08-wasm
|-- cosmos-in-evm
|   `-- ibc-solidity
|-- linea
|   `-- ibc-go-v8-08-wasm
|-- movement
|   `-- ibc-go-v8-08-wasm
|-- scroll
|   `-- ibc-go-v8-08-wasm
`-- tendermint
    `-- ibc-go-v8-08-wasm
```

grouping with the client type first makes it much easier to compare implementations, share test vectors, consolidate documentation, etc. this design also ensures that our libraries for implementing light clients are properly separated from the actual implementations.

benluelo:
duplicate of #3245

PR #3124, opened 2024-10-19
feat: split cometbls light client types out from unionlabs

benluelo:
- introduced `cometbls-light-client-types`
  - i put this under `light-clients/cometbls-light-client/types` for now, but we can pretty trivially move this down the line when we do https://github.com/unionlabs/union/issues/2997
  - also introduced a new `ChainId` type in this library to enforce the length invariant on cometbls chain ids, allowing us to remove several duplicated checks in various places.
  - used `alloy::sol!` for ethabi spec here instead of `ethers`
- slightly changed the constructor and error type of the `Bounded*` types to make conversion much easier
  - introduced `new_const` constructor for use in const contexts, this functions exactly the same as the old `new` constructor
- small change to `required!`, it now requires that the error type impls `From<MissingField>`
  - as such, derived `thiserror::Error` for some final remaining error types
- introduced `impl_proto_via_try_from_into!`
  - i want to eventually remove `macros::model`, since it doesn't quite work across many crates as well as it does when its designed for one crate (`unionlabs`). This covers the same boilerplate as the `#[model]` macro, but as a declarative macro.
- `unionlabs::hash::Hash<N>` <-> `alloy::core::primitives::FixedBytes<N>`
- remove logic inside of `canonicalize_stored_header` as this logic is no longer needed with the new evm ibc stack
  - i would remove it completely, but thats a larger change that should be done as we migrate each light client. it currently just passes through the original value, which is fine since the evm light clients don't even work right now anyways 😁
- cleaned up clippy warnings across the repo, they were driving me insane. i'll fix the clippy ci job later


aeryz:
Why don't we put `cometbls-light-client-types` under `lib`?

benluelo:
it's just where i put it, no specific reason. tbh i didn't think too much about it considering that we'll be moving a lot of this around soon anyways

light-clients/ethereum-light-client/src/client.rs
```diff
@@ -498,33 +492,33 @@ pub fn do_verify_membership(
...
-        >::decode_as::<Proto>(raw_value.as_ref())
-        .map_err(CanonicalizeStoredValueError::CometblsConsensusStateDecode)?
-        .0
-        .data
-        .encode_as::<EthAbi>(),
-        _ => raw_value,
-    };
-
-    Ok(canonical_value)
+    // let path = path
```
aeryz, line 498: Can we create an issue to fix the eth client. Let's first fix it and make sure it works, then we make a common lib for evm libs.

aeryz (approved):

light-clients/ethereum-light-client/src/client.rs
benluelo: https://github.com/unionlabs/union/issues/3127

PR #3125, opened 2024-10-19
Hubble contract filtering

qlp:


benluelo (approved):

hubble/src/main.rs
```diff
@@ -1,7 +1,4 @@
-#![feature(more_qualified_paths)]
-#![feature(try_blocks)]
-#![allow(clippy::needless_lifetimes, clippy::manual_async_fn)]
-#![feature(async_closure)]
```
benluelo, line 4: were we not actually using these features?

benluelo (approved):

hubble/src/main.rs
qlp: I think we only used them in the original indexers

PR #3130, opened 2024-10-21
feat: split arbitrum light client types out from unionlabs

benluelo:
- split arbitrum light client types out from unionlabs


lib/chain-utils/src/lib.rs
```diff
@@ -1,6 +1,6 @@
 #![feature(trait_alias)]
 
-pub mod arbitrum;
+// pub mod arbitrum;
```
benluelo, line 3: commented out for now but kept as it is currently unused but will be brought back & reused for the arbitrum voyager modules soon

cor (approved):

PR #3131, opened 2024-10-21
feat: split scroll light client types out from unionlabs

benluelo:
- split scroll light client types out from unionlabs
- change a few usages of other `*-light-client-types` types to import from the root re-export


lib/chain-utils/src/lib.rs
```diff
@@ -4,7 +4,7 @@
 pub mod berachain;
 pub mod cosmos;
 pub mod ethereum;
-pub mod scroll;
+// pub mod scroll;
```
qlp, line 7: delete?

hubble/Cargo.toml
```diff
@@ -48,6 +48,7 @@ protos                        = { workspace = true, features = ["client"] }
 regex                         = "1.10.5"
 reqwest                       = { workspace = true, features = ["json", "blocking"] }
 scroll-api                    = { workspace = true }
+scroll-light-client-types     = { version = "0.1.0", path = "../light-clients/scroll-light-client/types", features = ["proto"] }
```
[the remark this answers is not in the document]
benluelo: this actually needs to be fixed, it was added incorrectly

lib/chain-utils/src/lib.rs
benluelo: keeping for the same reason as i kept the arbitrum one https://github.com/unionlabs/union/pull/3130#discussion_r1808074718

qlp (approved):

qlp (approved):

PR #3152, opened 2024-10-24
feat: split out tm lc, bera lc, cometbft rpc, and galois rpc types

benluelo:


lib/galois-rpc/src/zero_knowledge_proof.rs
```diff
@@ -1,19 +1,15 @@
-use macros::model;
+use serde::{Deserialize, Serialize};
 
 // REVIEW: Are these fields fixed size?
-#[model(proto(raw(protos::union::galois::api::v3::ZeroKnowledgeProof), into, from))]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct ZeroKnowledgeProof {
     #[serde(with = "::serde_utils::hex_string")]
-    #[debug(wrap = ::serde_utils::fmt::DebugAsHex)]
```
qlp, line 7: seems useful? why remove this?

light-clients/arbitrum-light-client/types/src/header.rs
```diff
@@ -31,7 +31,7 @@ pub mod proto {
     impl_proto_via_try_from_into!(Header => protos::union::ibc::lightclients::arbitrum::v1::Header);
 
     impl TryFrom<protos::union::ibc::lightclients::arbitrum::v1::Header> for Header {
-        type Error = TryFromHeaderError;
```
qlp, line 34: just because you like this name better? any other reason?

lib/ibc-vm-rs/Cargo.toml
```diff
@@ -9,6 +9,7 @@ version      = "0.1.0"
 enumorph                 = { workspace = true }
 frame-support-procedural = { workspace = true }
 hex                      = { workspace = true }
+ibc-events               = { version = "0.1.0", path = "../ibc-events" }
```
qlp, line 12: why don't we add these to workspace?

lib/galois-rpc/src/zero_knowledge_proof.rs
benluelo: No longer have access to our custom debug macro here. This will be replaced with a custom bytes type soon

light-clients/arbitrum-light-client/types/src/header.rs
benluelo: Consistency with all other types. Intended usage is `<module name>::proto::Error`, it looks a lot cleaner as well and makes it easier to maintain a naming scheme

qlp (dismissed):
looking good, but ICE in build

lib/ibc-vm-rs/Cargo.toml
benluelo: ended up fixing this in this PR to trigger garnix

hussein-aitlahcen (approved):

qlp (approved):

PR #3184, opened 2024-10-31
upgrade protos

benluelo:


lib/cometbft-types/src/crypto/public_key.rs
```diff
@@ -1,13 +1,18 @@
...
 pub enum PublicKey {
-    Ed25519(#[serde(with = "::serde_utils::hex_string")] Vec<u8>),
-    Secp256k1(#[serde(with = "::serde_utils::hex_string")] Vec<u8>),
-    Bn254(#[serde(with = "::serde_utils::hex_string")] Vec<u8>),
-    Bls12_381(#[serde(with = "::serde_utils::hex_string")] Vec<u8>),
+    #[serde(rename = "tendermint/PubKeyEd25519")]
+    Ed25519(Bytes<Base64>),
+    #[serde(rename = "tendermint/PubKeySecp256k1")]
+    Secp256k1(Bytes<Base64>),
+    #[serde(rename = "cometbft/PubKeyBls12_381")]
```
aeryz, line 12: why this doesn't have tm alias but bn254 does?

aeryz (approved):

lib/cometbft-types/src/crypto/public_key.rs
benluelo: compatability with union-testnet-8: https://rpc.testnet-8.union.build/validators

image

PR #3195, opened 2024-11-01
feat: voyager event source for new evm ibc stack

benluelo:


lib/ibc-solidity/src/lib.rs
```diff
@@ -0,0 +1,629 @@
+pub mod ibc {
+    alloy::sol! {
+        // #![sol(cfg_attr(feature = "rpc", rpc))]
```
aeryz, line 3: do we need this?

lib/ibc-solidity/src/lib.rs
```diff
@@ -0,0 +1,629 @@
...
+            );
+            #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+            event ConnectionOpenConfirm(
+                uint32 connectionId,
+                uint32 clientId,
+                uint32 counterpartyClientId,
+                uint32 counterpartyConnectionId
+            );
+
+            // error ErrInvalidProof();
```
aeryz, line 163: remove plis

voyager/plugins/event-source/ethereum/src/main.rs
```diff
@@ -201,14 +198,15 @@ impl Module {
             .await
             .map_err(json_rpc_error_to_error_object)?;
 
+        let other_port_id = self_channel.counterparty.port_id.clone();
```
aeryz, line 201: nit `counterparty_*` for both

aeryz (approved):

lib/ibc-solidity/src/lib.rs
benluelo: i have to figure out a way to properly feature gate the rpc generation, its not currently possible which is super annoying

voyager/plugins/event-source/ethereum/src/main.rs
benluelo: i specifically use other_ as opposed to counterparty_ to avoid confusion in places where there is no "counterparty" (i.e. packets with src/dst)

lib/ibc-solidity/src/lib.rs
benluelo: will clean up this crate in a followup PR

PR #3203, opened 2024-11-05
Hubble reload block after finalization

qlp:


hubble/src/indexer/mod.rs
```diff
@@ -29,9 +29,33 @@ pub struct Indexer<T: FetcherClient> {
...
     pub context: T::Context,
 }
 
+#[derive(Clone, Debug, serde::Deserialize)]
+pub struct FinalizerConfig {
+    // how many blocks to wait until a block is considered finalized (ie. there should be no reorgs).
+    // compensates for height differences between rpcs
+    // default: 5
+    #[serde(default)]
+    pub delay_blocks: usize,
```
benluelo, line 42: this will default to 0, not 5 - you need to define a function and set `#[serde(default = "function_that_returns_default_value_you_want")]` and then use that same function in `FinalizerConfig::default` for consistency

benluelo: also good to use `NonZeroUsize` here probably?

hubble/src/indexer/mod.rs
```diff
@@ -29,9 +29,33 @@ pub struct Indexer<T: FetcherClient> {
...
+    // reload all block data after a block is considered finalized
+    // compensates for rpcs returning inconsistent results for non-finalized blocks.
+    // default: true
+    #[serde(default)]
+    pub reload: bool,
```
benluelo, line 47: this will default to false, requires the same as my previous comment

hubble/src/indexer/mod.rs
qlp: fixed the default. didn't do non-zero, as zero would be good value if we have an rpc that is consistent

hubble/src/indexer/mod.rs
qlp: done

hubble/src/indexer/mod.rs
```diff
@@ -38,20 +38,30 @@ pub struct FinalizerConfig {
...
 }
 
+impl FinalizerConfig {
+    pub fn finalizer_config_default_delay_blocks() -> usize {
+        5
+    }
+
+    pub fn finalizer_config_default_reload() -> bool {
+        true
+    }
```
benluelo, line 57: minor nit: no need to prefix these functions with `finalizer_config_*`

benluelo (approved):

hubble/src/indexer/mod.rs
qlp: fixed (and squashed)

benluelo (approved):

PR #3235, opened 2024-11-08
fix(site): disable retina

cor:


cor:
Upon extensive profiling and benchmarking, we concluded that it makes no difference in performance regardless of browser. It only seems to reduce visual quality

PR #3248, opened 2024-11-12
feat(union-ibc): add `GetChannel` and `GetChannels` queries

hussein-aitlahcen:


cosmwasm/union-ibc/src/contract.rs
```diff
@@ -1477,6 +1480,25 @@ fn save_connection(
...
+    deps: DepsMut,
+    owner: Addr,
+    channel_id: u32,
+    channel: &Channel,
+) -> Result<(), ContractError> {
+    CHANNEL_OWNER.save(deps.storage, channel_id, &owner)?;
+    CONTRACT_CHANNELS.update(deps.storage, owner, |v| -> Result<_, ContractError> {
+        Ok(match v {
+            Some(mut set) => {
+                set.insert(channel_id);
```
benluelo, line 1493: what happens if the channel already exists? should check if the call to insert returns `Some`

hussein-aitlahcen: it's impossible as the `create` variant is only called on `init` and `try`

benluelo (approved):

benluelo (approved):

PR #3255, opened 2024-11-13
feat: migrate to svelte 5

o-az:
the first commit is the result of running `npx sv@latest migrate svelte-5` without any manual changes.


o-az:
update: this is shelved until tanstack query, tanstack table and shadcn-svelte are all svelte-5 ready and stable.

PR #3257, opened 2024-11-13
feat(ucs01): plug-in union-ibc

hussein-aitlahcen:
- introduce the `union-ibc` callbacks
- slightly abstract the protocol traits to be agnostic w.r.t to the underlying ibc impl
- coerce union-ibc types in legacy ibc (sequence 0 for packets etc..)
- extract union-ibc msgs/queries to `union-ibc-msg`


cosmwasm/union-ibc-msg/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+license-file = { workspace = true }
+name         = "union-ibc-msg"
+repository   = "https://github.com/unionlabs/union"
+version      = "1.0.0"
+
+[lints]
+workspace = true
+
+[lib]
+crate-type = ["cdylib", "rlib"]
```
benluelo, line 13: can be removed

aeryz (dismissed):

aeryz (approved):

PR #3274, opened 2024-11-18
fix(libwasmvm): use rust 1.77

hussein-aitlahcen:


cor:
Solved by changing linker args, PR no longer relevant

PR #3280, opened 2024-11-19
fix(evm): use `wasm` key and prefixes in cometbls client

hussein-aitlahcen:


evm/contracts/core/24-host/IBCStore.sol
```diff
@@ -5,12 +5,11 @@ import "../05-port/IIBCModule.sol";
 import "../Types.sol";
 
 library IBCStoreLib {
-    bytes32 public constant COMMITMENT_PREFIX = keccak256("ethibc");
+    bytes32 public constant COMMITMENT_PREFIX = keccak256("wasm");
```
benluelo, line 8: its not keccak(wasm) right? just wasm?

hussein-aitlahcen: yep good catch!

PR #3285, opened 2024-11-21
feat: replace cspell with typos

o-az:
typos reference: https://github.com/crate-ci/typos/blob/master/docs/reference.md

I think we can also delete dictionary.txt btw


cor (approved):
this will save 10.000 hours

PR #3287, opened 2024-11-21
chore(uniond): update proto generation and open API spec

PoisonPhang:


tools/rust-proto.nix
```diff
@@ -598,7 +599,7 @@ _: {
               name = name + "-protos";
               path = src;
             }
-          ) all-protos-to-build
+          ) (dbg all-protos-to-build)
```
benluelo, line 602: Remove dbg?

benluelo (dismissed):

PR #3294, opened 2024-11-23
fix: graphiql light mode theme

o-az:


cor:
image
seems still dark to me

o-az:
> image seems still dark to me

fixed

PR #3334, opened 2024-12-04
feat: new transfer flow

o-az:
the work is in `app/src/routes/transfer-new` until the transfer form is fully refactored then just `app/src/routes/transfer`

purpose of this pr is to complete the first stage of the intent process; namely: validation step. Once the form is filled and the transfer is submitted, the validation results are logged


app/src/routes/transfer-new/+page.svelte
```diff
@@ -0,0 +1,306 @@
...
+import { transferSchema, type TransferSchema } from "./validation.ts"
+import { getCosmosChainInfo } from "$lib/wallet/cosmos/chain-info.ts"
+import { submittedTransfers } from "$lib/stores/submitted-transfers.ts"
+import { parseUnits, formatUnits, type HttpTransport, getAddress } from "viem"
+import { aptosStore, userAddressAptos, getAptosWallet } from "$lib/wallet/aptos"
+import { cosmosStore, getCosmosOfflineSigner } from "$/lib/wallet/cosmos/config.ts"
+import { type Writable, writable, derived, get, type Readable } from "svelte/store"
+import { custom, switchChain, getConnectorClient, waitForTransactionReceipt } from "@wagmi/core"
+
+type SearchParams = { [key: string]: string }
```
cor, line 54: This should have a stronger type, rather than just key:value

app/src/routes/transfer-new/+page.svelte
```diff
@@ -0,0 +1,306 @@
...
+
+let transferQueryOptions = queryOptions<SearchParams>({
+  enabled: query => false,
+  placeholderData: keepPreviousData,
+  queryKey: ["transfer-state", $page.url.toString()],
+  initialData: Object.fromEntries($page.url.searchParams),
+  queryFn: ({ queryKey, signal, meta }) => Object.fromEntries($page.url.searchParams)
+})
+
+let state = createQuery(transferQueryOptions)
```
cor, line 76: `state` is not really descriptive

app/src/routes/transfer-new/+page.svelte
```diff
@@ -0,0 +1,306 @@
...
+let userAddress = derived(
+  [userAddrCosmos, userAddrEvm, userAddressAptos],
+  ([$cosmos, $evm, $aptos]) => ({ evm: $evm, aptos: $aptos, cosmos: $cosmos })
+)
+
+$: asset = $page.url.searchParams.get("asset") || $state.data?.asset
+$: amount = $page.url.searchParams.get("amount") || $state.data?.amount
+$: receiver = $page.url.searchParams.get("receiver") || $state.data?.receiver
+$: source = $page.url.searchParams.get("source") || $state.data?.source
+$: destination = $page.url.searchParams.get("destination") || $state.data?.destination
```
cor, line 92: What do tehse lines do? it seems that currently `asset` is defined in three places:

1. `searchParams.get("asset")`
2. `searchParams.data?.asset`
3. `$: asset = ...`

Let's make sure we have only 1, strongly-typed, place where de fine the user input that is to be validated.

If we need to initially set this manually based on some initial data from the URL, I suggest handling this in an onMount call rather than introducing reactive state

app/src/routes/transfer-new/+page.svelte
```diff
@@ -0,0 +1,306 @@
...
+
+$: sourceChain = writable(chains.find(chain => chain.chain_id === source))
+$: destinationChain = writable(chains.find(chain => chain.chain_id === destination))
```
cor, line 95: This is derived info that should never be written to. Let's use a `let destinationChain = derived(...)` which is of type `Readable<T>` rather than `Writable<T>`

app/src/routes/transfer-new/+page.svelte
```diff
@@ -0,0 +1,306 @@
...
+$: rawBalances = userBalancesQuery({
+  chains,
+  connected: true,
+  userAddr: $userAddress
+})
+
+$: balances = derived([rawBalances, sourceChain], ([$rawBalances, $sourceChain]) => {
+  if (!($sourceChain && $rawBalances)) return []
+  return $rawBalances.data[source] ?? []
+})
```
cor, line 106: balances is derived so it should not have `$:`

app/src/routes/transfer-new/+page.svelte
```diff
@@ -0,0 +1,306 @@
...
+
+// @ts-ignore
+$: assetInfo = $balances.find(x => x?.address === asset)
+$: {
+  console.log(asset, assetInfo)
+}
```
cor, line 112: this is derive data

cor (changes requested):
I like the general structure of encoding pre-validated user input in the URL, but I feel that the data structures need some work.

1. For all state, let's ensure that it is defined at most once, in strongly-typed structures
2. let's refrain from `$:` as it is fully dropped in svelte 5,
3. derived state should never be writable

Also, I noticed that whenever I click a form element which opens a dialog, then exit the dialog, then click one again, nothing happens. I need to click everything twice

Excited about the direction where this is heading!

PR #3348, opened 2024-12-06
feat(move-bindgen): properly support generics

benluelo:


lib/move-bindgen-derive/src/lib.rs
```diff
@@ -36,23 +35,23 @@ fn derive_type_tagged(
...
+
+    //         assert!(mnv.path.is_ident("module"));
+    //         match &mnv.value {
+    //             Expr::Path(expr) => expr.path.require_ident().unwrap().clone(),
+    //             _ => panic!("???"),
+    //         }
+    //     })
+    //     .collect::<Vec<_>>()
+    //     .try_into()
+    //     .unwrap();
```
hussein-aitlahcen, line 54: intented to be keeped?

hussein-aitlahcen (approved):

lib/move-bindgen-derive/src/lib.rs
benluelo: yeah I need to figure out what to do with this

PR #3357, opened 2024-12-10
feat(berachain): make light client L2 to comet L1

hussein-aitlahcen:


cosmwasm/union-ibc/light-clients/tendermint/src/client.rs
```diff
@@ -105,13 +106,25 @@ impl IbcClient for TendermintLightClient {
...
+            )?)
+        } else {
+            Ok(verify_header(
+                client_state,
+                consensus_state,
+                header,
+                ctx.env.block.time,
+                &SignatureVerifier::new(Ed25519Verifier::new(ctx.deps)),
+            )?)
+        }
```
benluelo, line 127: why not match on the key type instead of if-let?

lib/berachain-light-client-types/src/client_state.rs
```diff
@@ -1,12 +1,11 @@
...
 #[derive(Debug, Clone, PartialEq)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct ClientState {
-    pub comet: tendermint_light_client_types::ClientState,
-    pub execution_chain_id: U256,
-    // latest execution height
-    pub execution_latest_height: u64,
+    pub l1_client_id: ClientId,
+    pub chain_id: U256,
+    pub latest_height: u64,
```
benluelo, line 8: perhaps add a comment explaining that this is the latest height of the eth execution layer?

benluelo (approved):

PR #3374, opened 2024-12-13
Docs improvements

tudorpintea999:
Hello
I did some small fixes on the docs
It not much work but it s honest work
Hope it helps.
Br,
Tudor


benluelo:
thank you for opening the pull request. however, this touches many files and has some changes that are incorrect. closing for those reasons and also because this has gone stale.

PR #3375, opened 2024-12-13
fix(evm): typo in event

hussein-aitlahcen:


evm/contracts/core/04-channel/IBCPacket.sol
```diff
@@ -14,7 +14,7 @@ library IBCPacketLib {
     bytes32 public constant COMMITMENT_NULL = bytes32(uint256(0));
 
     event SendPacket(IBCPacket packet);
-    event RecvPacket(IBCPacket packets, address maker, bytes makerMsg);
+    event RecvPacket(IBCPacket packet, address maker, bytes makerMsg);
```
benluelo, line 17: fix in lib/ibc-solidity as well?

benluelo (approved):

evm/contracts/core/04-channel/IBCPacket.sol
hussein-aitlahcen: It's fixed already lol?

benluelo (approved):

Issue #3363, opened 2024-12-11
use unionlabs::hash::H256 in evm-storage-verifier


benluelo:
remove primitive-types usage in this crate and use the types from unionlabs instead: https://github.com/unionlabs/union/blob/main/lib/evm-storage-verifier/Cargo.toml#L12

this *should* be a pretty trivial swap, since `unionlabs::hash::Hash` implements rlp encoding: https://github.com/unionlabs/union/blob/e652dd1905e77d0138fb68b351cb17248d830689/lib/unionlabs/src/hash.rs#L567-L586

PR #3377, opened 2024-12-13
Refactor: Use h256 from own crate in evm storage verifier

atahanyild:
Fixes https://github.com/unionlabs/union/issues/3363

Since Unionlabs::Hash::H256 is already implemented, this refactors the codes that uses primitive_types::H256 to our own one.

Additionally ,this PR implements AsMut to hash.rs as Hasher need AsMut trait to be resolved.


benluelo (approved):
thanks for the pr! lgtm 🚀

PR #3382, opened 2024-12-16
add a missing feature in dependency

larry0x:
for the ibc-union-spec crate, you need to enable the `serde` feature on the ibc-solidity dependency, in order to use the solidity types in cosmwasm messages.


benluelo:
closing as no longer relevant, ibc-union-spec now defines it's own types (as well as a compatability feature with ibc-solidity)

PR #3414, opened 2024-12-30
feat(voyager): replace u64 timestamp with Timestamp type across modules

### LINKED ISSUE #3388: Introduce newtype for timestamp in voyager
We currently use a raw `i64` or `u64` for timestamps throughout voyager:

https://github.com/unionlabs/union/blob/main/lib/voyager-message/src/module.rs#L398-L399
https://github.com/unionlabs/union/blob/main/lib/voyager-core/src/lib.rs#L275-L277

...etc

This should be replaced with a newtype that wraps a u64, storing the timestamp as nanoseconds. This type should live in `voyager-core`.

atahanyild:
closes #3388

This PR implements new Timestamp type under voyager core, unifying types that express time as nanoseconds

### DIFF
--- a/lib/voyager-message/src/rpc.rs
+++ b/lib/voyager-message/src/rpc.rs
@@ -8,7 +8,7 @@ use macros::model;
 use serde::de::DeserializeOwned;
 use serde_json::{json, Value};
 use unionlabs::{bytes::Bytes, ibc::core::client::height::Height, ErrorReporter};
-use voyager_core::IbcSpecId;
+use voyager_core::{IbcSpecId, Timestamp};
 
 use crate::{
     context::LoadedModulesInfo,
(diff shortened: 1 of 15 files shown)


lib/voyager-message/src/rpc.rs
```diff
@@ -39,7 +39,11 @@ pub trait VoyagerRpc {
 
     #[method(name = "queryLatestTimestamp")]
     // TODO: Make this return a better type than i64
```
benluelo, line 41: nit: remove the comment

voyager/modules/consensus/movement/src/main.rs
```diff
@@ -145,7 +145,11 @@ impl ConsensusModuleServer for Module {
 
     /// Query the latest finalized timestamp of this chain.
     // TODO: Make this return a better type than i64
```
benluelo, line 147: nit: remove comment

voyager/modules/consensus/tendermint/src/main.rs
```diff
@@ -122,7 +122,11 @@ impl ConsensusModuleServer for Module {
     /// Query the latest finalized timestamp of this chain.
     // TODO: Use a better timestamp type here
```
benluelo, line 123: also this comment

benluelo (approved):

benluelo (approved):

PR #3434, opened 2025-01-02
feat(helix): configure vscode-solidity-server

cor:
Language server provided by `solc` has long been broken, so I used
https://www.npmjs.com/package/vscode-solidity-server .

I tried packaging it, but node-gyp blocked me: https://github.com/NixOS/nixpkgs/issues/370370.
Config in this PR works, but does require you to run `npm install` in `evm/` once to get the solidity lsp.

### DIFF
--- a/evm/package-lock.json
+++ b/evm/package-lock.json
@@ -1,0 +1,6643 @@
+{
+  "name": "evm dev tools",
+  "version": "0.0.1",
+  "lockfileVersion": 3,
+  "requires": true,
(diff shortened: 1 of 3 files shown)


evm/package-lock.json
benluelo, line 1: disgusting

benluelo (approved):

evm/package-lock.json
cor: i agree

PR #3439, opened 2025-01-06
chore: port solidity ibc-union tests to cosmwasm

PoisonPhang:
Port solidity tests to CosmWasm

- `02-client`
- `03-connections`
- `04-channels`

Missing Tests:
- invalid proof tests
- commitment stored (04-channel/packet)
- async ack on batch ack
- ok_n tests


aeryz (approved):
zklgtm!

hussein-aitlahcen (approved):

PR #3456, opened 2025-01-08
Cosmos <> EVM  state lens

aeryz:


lib/cometbls-light-client-types/src/client_state.rs
benluelo, line 1: revert changes?

lib/tendermint-light-client-types/src/client_state.rs
```diff
@@ -95,6 +102,8 @@ pub mod proto {
                     .collect::<Result<Vec<_>, _>>()
                     .map_err(Error::ProofSpecs)?,
                 upgrade_path: value.upgrade_path,
+                // // FIXME: we need to define the tm proto ourself and regenerate
```
benluelo, line 105: change comment to instead mention that the contract address field is not used for the ibc-go native implementation

lib/unionlabs/src/ibc/lightclients/evm_in_cosmos/header.rs
```diff
@@ -1,74 +1 @@
-use macros::model;
```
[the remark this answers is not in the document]
aeryz: moved to its own lib like the other clients, removing this

benluelo (approved):

PR #3508, opened 2025-01-14
feat(ibc-union): Add query responses 

adairrr:
Adds `QueryResponses` macro for the `union-ibc` contract along with a cw-orch interface. 

Current issues:
- Some return types do not implement cw-serde and need to 
- We should feature-gate the cw-orch interfaces.


cosmwasm/ibc-union/core/src/interface.rs
```diff
@@ -0,0 +1,31 @@
+pub const UNION_IBC: &str = "union:union-ibc";
+
+#[cw_orch::interface(
+    crate::msg::InstantiateMsg,
+    crate::msg::ExecuteMsg,
+    crate::msg::QueryMsg,
+    crate::msg::MigrateMsg,
+    id = EVM_NOTE
+)]
+pub struct UnionIbc<Chain>;
```
benluelo, line 10: `IbcUnion`

lib/unionlabs-primitives/src/fixed_bytes.rs
```diff
@@ -236,6 +236,32 @@ impl<'de, const BYTES: usize, E: Encoding> serde::Deserialize<'de> for FixedByte
     }
 }
 
+#[cfg(feature = "schemars")]
+impl<const BYTES: usize, E: Encoding> schemars::JsonSchema for FixedBytes<BYTES, E> {
+    fn schema_name() -> String {
+        "Hash".to_string()
+    }
+
+    /// The `FixedBytes` object is serialized as an array in JSON
```
benluelo, line 245: it's serialized as string according to `E` (hex, hex unprefixed, base64, etc). jsonschema for FixedBytes and Bytes are a bit more complicated if you want full correctness, but for now just make them be strings

cosmwasm/ibc-union/core/src/interface.rs
```diff
@@ -0,0 +1,31 @@
+pub const UNION_IBC: &str = "union:union-ibc";
+
+#[cw_orch::interface(
+    crate::msg::InstantiateMsg,
+    crate::msg::ExecuteMsg,
+    crate::msg::QueryMsg,
+    crate::msg::MigrateMsg,
+    id = EVM_NOTE
```
benluelo, line 8: what is `EVM_NOTE`?

adairrr: Updated to `UNION_IBC`

PoisonPhang:
Replaced by #4584

PR #3531, opened 2025-01-16
feat(voyager): make the state lens client update module fully generic

benluelo:
internal voyager change to add the client id of the client being updated to the update client request message, such that update plugins can inspect the client state for possible client-specific configuration options.


hussein-aitlahcen (approved):

hknnr:
)

PR #3601, opened 2025-01-21
Show both graphql and onchain data for assets

cor:
- **feat(app): fetch onchain asset info**
- **feat(app): show both graphql and onchain token info**
- **chore: fmt**


gulsenmausika:
hey

PR #3631, opened 2025-01-24
chore(zkgm): virtualize tokenfactory

aeryz:
Intended behaviour:
1. `transfer`:
    a. **Minter is TF, sent wrapped token**: Call `TokenFactoryMsg::BurnFrom` -> tf minter checks whether the funds are provided and burns.
    b. **Minter is CW20, sent wrapped token**: Call `TokenFactoryMsg::BurnFrom` -> cw20 minter just burns the tokens.
    c. **Minter is TF, sent native token**: Call `LocalTokenMsg::TakeFunds` -> tf minter checks whether the funds are provided and noops.
    d. **Minter is CW20, sent CW20**: Call `LocalTokenMsg::TakeFunds` -> cw20 minter calls `CW20::TransferFrom`
    e. **Minter is CW20, sent native token**: Call `LocalTokenMsg::TakeFunds` -> cw20 minter sees funds are provided, saves the given `denom` as native and noops.

2. `recv_transfer`:
    a. **Minter is TF**: Nothing fancy, just delegate the messages to `TF`.
    b. **Minter is CW20, need to create a wrapped token**:
        1. Call `TokenFactoryMsg::CreateDenom`: Since `CW20` doesn't have the token info yet, it just noops.
        2. Call `TokenFactoryMsg::SetDenomMetadata`: `CW20` now has the token info, so it instantiates a contract. To save the denom to the denom to addr mapping, it saves the denom to `DENOM_TO_BE_STORED`. NOTE THAT THIS IS A SINGLE VALUE STORAGE.
        3. Step 2 results in `reply` if success, in the reply, read `DENOM_TO_BE_STORED` and save denom and addr to `DENOM_TO_ADDR` mapping.
        4. Call `TokenFactoryMsg::MintTokens`:  Read from `DENOM_TO_ADDR` mapping and mint the tokens.

    c. **Minter is CW20, need to unescrow local token**: CW20 just calls `CW20::Transfer`
    d. **Minter is CW20, need to unescrow native token**: CW20 sees that the given denom previously been saved as a native token. So it calls `BankMsg::Send`.


cosmwasm/cw20-token-minter/src/contract.rs
```diff
@@ -0,0 +1,291 @@
...
+use ucs03_zkgm_token_minter_api::{
+    BaseTokenResponse, ExecuteMsg, LocalTokenMsg, MetadataResponse, QueryMsg, WrappedTokenMsg,
+};
+
+use crate::{
+    error::Error,
+    state::{Config, ADDR_TO_DENOM, CONFIG, DENOM_TO_ADDR, DENOM_TO_BE_STORED},
+};
+
+pub const NATIVE_TOKEN_STORE_PREFIX: u32 = 0x1;
```
benluelo, line 17: unused?

cosmwasm/cw20-token-minter/src/state.rs
```diff
@@ -0,0 +1,17 @@
...
+    pub cw20_code_id: u64,
+}
+
+pub const CONFIG: Item<Config> = Item::new("conf");
+
+pub const DENOM_TO_ADDR: Map<String, Addr> = Map::new("dta");
+
+pub const ADDR_TO_DENOM: Map<Addr, String> = Map::new("atd");
+
+pub const DENOM_TO_BE_STORED: Item<String> = Item::new("dtbs");
```
benluelo, line 17: docs on these would be good

cosmwasm/cw20-token-minter/src/error.rs
```diff
@@ -0,0 +1,32 @@
...
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error(transparent)]
+    StdError(#[from] StdError),
+
+    #[error("only admin can execute")]
+    OnlyAdmin,
+
+    #[error("this contract cant mint the token with denom {0} because it's not the owner")]
```
benluelo, line 12:
```suggestion
    #[error("this contract can't mint the token with denom {0} because it's not the owner")]
```

PR #3639, opened 2025-01-24
feat: move Slot type into it's own library

### LINKED ISSUE #3417: Move `Slot` type into it's own library
The type in this file: https://github.com/unionlabs/union/blob/main/lib/unionlabs/src/ethereum/slot.rs

could be moved into it's own library. It would be good to also create a small binary to help calculate slots given a storage layout.

cozfuttu:
Closes #3417 

Created a small binary for slot calculation as well. Usage:
`cargo run --bin slot-calculator --layout "mapping(uint256 => mapping(uint256 => uint256)[])" --keys 100 1 123`

### DIFF
--- a/tools/slot-calculator/src/main.rs
+++ b/tools/slot-calculator/src/main.rs
@@ -1,0 +1,206 @@
+use std::{collections::VecDeque, str::FromStr, vec};
+
+use clap::{Arg, Command};
+use slotlib::{MappingKey, Slot};
+use typed_arena::Arena;
+use unionlabs::primitives::{H256, U256};
+
+// Examples:
+// mapping(uint256 => uint256) => ["uint256"]
+// mapping(uint256 => uint256[]) => ["uint256", "uint256[]"]
+// mapping(uint256 => mapping(uint256 => uint256)) => ["uint256", "uint256"]
+// mapping(uint256 => mapping(uint256 => mapping(uint256 => uint256)[])[]) => ["uint256", "mapping[]", "uint256", "mapping[]", "uint256"]
+// mapping(uint256 => mapping(uint256 => mapping(uint256 => mapping(uint256 => uint256)[]))[]) => ["uint256", "mapping[]", "uint256", "uint256", "mapping[]", "uint256"]
+// mapping(uint256 => mapping(uint256 => uint256)[]) => ["uint256", "mapping[]", "uint256"]
+fn parse_layout(layout: &str) -> Result<Vec<String>, String> {
+    // 1. Basic validation
+    if !layout.starts_with("mapping(") || !layout.ends_with(')') {
+        return Err("Invalid layout".to_string());
+    }
+
+    // 2. Remove all 'mapping', then strip the first '(' and last ')' character.
+    let mut sanitized = layout.replace("mapping", "");
+    sanitized = sanitized[1..sanitized.len() - 1].to_string();
+
+    // 3. Split by '=>', trim and remove any leftover '('
+    let mut split_by_arrow: Vec<String> = sanitized
+        .split("=>")
+        .map(|s| s.trim().replace("(", ""))
+        .collect();
+
+    // cloning the array so we can iterate indexes from the *original* split_by_arrow
+    // while we modify the *current* split_by_arrow in place.
+    let original_split = split_by_arrow.clone();
+
+    for (i, element) in original_split.iter().enumerate() {
+        let mapping_array_count = element.matches(")[]").count();
+        if mapping_array_count == 0 {
+            continue;
+        }
+
+        // 1) split element by ")[]"
+        // 2) skip the first chunk
+        // 3) in each chunk, count how many times ')' appears, subtract 1
+        // 4) reverse at the end
+        let mut skip_indexes: Vec<usize> = element
+            .split(")[]")
+            .skip(1)
+            .map(|chunk| chunk.matches(')').count())
+            .collect();
+        skip_indexes.reverse();
+
+        // For each occurrence, insert "mapping[]" at index (i - j - skipCount).
+        for j in 1..=mapping_array_count {
+            let skip_count = skip_indexes[j - 1];
+            let insert_index = (i as usize) - (j as usize) - (skip_count as usize);
+            split_by_arrow.insert(insert_index as usize, "mapping[]".to_string());
+        }
+    }
+
+    // 4. Final check on the last element:
+    // if it doesn't have "[]" or if it still contains ")[]", pop it
+    if let Some(last) = split_by_arrow.last() {
+        if !last.contains("[]") || last.contains(")[]") {
+            split_by_arrow.pop();
+        }
+    }
+
+    Ok(split_by_arrow)
+}
+
+// MappingKey + U32
+#[derive(Debug)]
+pub enum KeyTypes<'a> {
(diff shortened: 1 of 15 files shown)


tools/slot-calculator/src/main.rs
```diff
@@ -0,0 +1,206 @@
...
+// Examples:
+// mapping(uint256 => uint256) => ["uint256"]
+// mapping(uint256 => uint256[]) => ["uint256", "uint256[]"]
+// mapping(uint256 => mapping(uint256 => uint256)) => ["uint256", "uint256"]
+// mapping(uint256 => mapping(uint256 => mapping(uint256 => uint256)[])[]) => ["uint256", "mapping[]", "uint256", "mapping[]", "uint256"]
+// mapping(uint256 => mapping(uint256 => mapping(uint256 => mapping(uint256 => uint256)[]))[]) => ["uint256", "mapping[]", "uint256", "uint256", "mapping[]", "uint256"]
+// mapping(uint256 => mapping(uint256 => uint256)[]) => ["uint256", "mapping[]", "uint256"]
+fn parse_layout(layout: &str) -> Result<Vec<String>, String> {
+    // 1. Basic validation
+    if !layout.starts_with("mapping(") || !layout.ends_with(')') {
```
aeryz, line 17: don't we also need to be able to calculate when the layout is not mapping but an array?

tools/slot-calculator/src/main.rs
```diff
@@ -0,0 +1,206 @@
...
+use unionlabs::primitives::{H256, U256};
+
+// Examples:
+// mapping(uint256 => uint256) => ["uint256"]
+// mapping(uint256 => uint256[]) => ["uint256", "uint256[]"]
+// mapping(uint256 => mapping(uint256 => uint256)) => ["uint256", "uint256"]
+// mapping(uint256 => mapping(uint256 => mapping(uint256 => uint256)[])[]) => ["uint256", "mapping[]", "uint256", "mapping[]", "uint256"]
+// mapping(uint256 => mapping(uint256 => mapping(uint256 => mapping(uint256 => uint256)[]))[]) => ["uint256", "mapping[]", "uint256", "uint256", "mapping[]", "uint256"]
+// mapping(uint256 => mapping(uint256 => uint256)[]) => ["uint256", "mapping[]", "uint256"]
+fn parse_layout(layout: &str) -> Result<Vec<String>, String> {
```
aeryz, line 15: @benluelo Do you know of a simple parsing lib for rust where we can define the grammar and do the parsing? It would be cleaner and safer to use a well tested parser.

benluelo: I highly recommend peg for this usecase, it makes it very trivial to define recursive grammars: https://docs.rs/peg/latest/peg/

tools/slot-calculator/src/main.rs
cozfuttu: ok, I'll add support for arrays as well.

tools/slot-calculator/src/main.rs
cozfuttu: i'll give it a shot, thanks for the feedback guys

tools/slot-calculator/src/main.rs
```diff
@@ -0,0 +1,152 @@
...
+            _ => return Err("Unsupported type".to_string()),
+        },
+        _ => return Err("Unsupported item".to_string()),
+    }
+}
+
+fn parse_mapping_key<'a>(key_type: &'a Type, key: &'a str) -> MappingKey<'a> {
+    match key_type {
+        Type::Uint(_, size) => {
+            let size = size.and_then(|s| Some(s.get())).unwrap_or(256);
```
benluelo, line 41: instead of .and_then with Some(), just use .map

tools/slot-calculator/src/main.rs
```diff
@@ -0,0 +1,152 @@
...
+            match size {
+                256 => MappingKey::Uint256(U256::from(
+                    key.parse::<u64>().expect("Invalid uint256 key"),
```
benluelo, line 44: parsing as u64 when this is a u256?

tools/slot-calculator/src/main.rs
```diff
@@ -0,0 +1,152 @@
...
+        eprintln!(
+            "Warning: Unused keys: {:?}. The calculated slot might be wrong. Please check the layout and keys you provided.",
+            split_keys
+        );
+    }
+    slot.slot()
+}
+
+fn main() {
+    let matches = Command::new("Slot Calculator (post-order)")
```
benluelo, line 100: use clap derive instead of the builder api

tools/slot-calculator/src/main.rs
```diff
@@ -0,0 +1,152 @@
...
+    let parsed_layout = parse2(parsed_layout).map_err(|e| e.to_string())?;
+
+    match &parsed_layout.items[0] {
+        Item::Variable(var) => match &var.ty {
+            Type::Mapping(map) => Ok(Type::Mapping(map.clone())),
+            Type::Array(arr) => Ok(Type::Array(arr.clone())),
+            _ => return Err("Unsupported type".to_string()),
+        },
+        _ => return Err("Unsupported item".to_string()),
+    }
```
benluelo, line 35: why not just parse directly into `Type`?

benluelo: then you could skip the visibility & name logic entirely

tools/slot-calculator/src/main.rs
```diff
@@ -0,0 +1,152 @@
...
+        None => return arena.alloc(Slot::Offset(U256::from(0u32))),
+    };
+    match parsed_layout {
+        Type::Mapping(mapping) => arena.alloc(Slot::Mapping(
+            build_slot(mapping.value.as_ref(), keys, arena),
+            parse_mapping_key(mapping.key.as_ref(), key),
+        )),
+        Type::Array(arr) => arena.alloc(Slot::Array(
+            build_slot(arr.ty.as_ref(), keys, arena),
+            U256::from(key.parse::<u64>().expect("Invalid array index")),
```
benluelo, line 78: parse as u256

tools/slot-calculator/src/main.rs
```diff
@@ -0,0 +1,152 @@
...
+}
+
+fn build_slot<'a>(
+    parsed_layout: &'a Type,
+    keys: &mut VecDeque<&'a str>,
+    arena: &'a Arena<Slot<'a>>,
+) -> &'a Slot<'a> {
+    let key: &str = match keys.pop_front() {
+        Some(k) => k,
+        None => return arena.alloc(Slot::Offset(U256::from(0u32))),
```
benluelo, line 69: what's the reasoning behind defaulting to a 0 slot here?

tools/slot-calculator/src/main.rs
```diff
@@ -0,0 +1,152 @@
...
+    let mut split_keys: VecDeque<&str> = keys.split(" ").collect();
+
+    let arena = Arena::new();
+    let slot = build_slot(&parsed_layout, &mut split_keys, &arena);
+    if split_keys.len() != 0 {
+        eprintln!(
+            "Warning: Unused keys: {:?}. The calculated slot might be wrong. Please check the layout and keys you provided.",
+            split_keys
+        );
+    }
```
benluelo, line 95: this should be a hard error

PR #3816, opened 2025-02-20
CW20 multibalance

cor:
Simple contract that queries multiple CW20 balances on a chain. Useful for reducing the amount of RPC calls from the app.

- **feat(cw20-multibalance): init**
- **feat(cw20-multibalance): fetch balances**


cor:
this ended up not yielding significant benefit as the max amount of queries was about ~2-3

PR #3838, opened 2025-02-21
Hide and query

Swepool:


app/src/lib/components/transfers-table/transfers-table.svelte
```diff
@@ -26,11 +26,17 @@ const transfersDataStore = derived([transfers, page], ([$transfers, $page]) => {
...
+
+        return (
           enabledChainIds.includes(transfer.source.chainId) &&
-          enabledChainIds.includes(transfer.destination.chainId)
-      )
+          enabledChainIds.includes(transfer.destination.chainId) &&
+          !Number.isNaN(parsedAmount) &&
+          parsedAmount !== 0
+        )
+      })
```
cor, line 39: We should not do this in the rendering portion, but rather in the GraphQL query

cor (changes requested):

cor (approved):

PR #3842, opened 2025-02-24
migrate zkgm and token minter

aeryz:


cosmwasm/ibc-union/app/ucs03-zkgm/src/contract.rs
```diff
@@ -1155,7 +1157,17 @@ pub fn migrate(
...
             Ok((res, None))
         },
-        |_deps, _migrate_msg, _current_version| Ok((Response::default(), None)),
+        |deps, migrate_msg, _current_version| {
+            let token_minter = TOKEN_MINTER.load(deps.storage)?;
+            Ok((
+                Response::default().add_message(WasmMsg::Migrate {
+                    contract_addr: token_minter.to_string(),
+                    new_code_id: migrate_msg.token_minter_code_id,
+                    msg: to_json_binary(&cosmwasm_std::Empty {})?,
```
benluelo, line 1166: if we're keeping this migration functionality on zkgm we should thread the message as well

benluelo (approved):

PR #3850, opened 2025-02-24
feat(zkgm): enforce decimals

hussein-aitlahcen:


aeryz (approved):

aeryz (commented):
If we are ditching the support for fungible asset v0, let's remove the data type and the corresponding encode/decode functions

PR #3902, opened 2025-03-01
feat(wasm): reduce size of built wasm blobs

benluelo:
this actually reduces the size of all rust builds, but the size improvements aren't as important for non-wasm builds.

### DIFF
--- a/tools/rust/crane.nix
+++ b/tools/rust/crane.nix
@@ -363,6 +363,18 @@
                   " -Z build-std=std,panic_abort -Z build-std-features=panic_immediate_abort --target ${buildStdTarget}"
                 );
               RUSTFLAGS = rustflags;
+
+              preBuild = ''
+                echo "cargoVendorDir: ${crateAttrs.cargoVendorDir}"
+                echo "rustToolchain: ${cargoBuildRustToolchain'}"
+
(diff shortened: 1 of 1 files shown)


tools/rust/crane.nix
```diff
@@ -363,6 +363,18 @@
                   " -Z build-std=std,panic_abort -Z build-std-features=panic_immediate_abort --target ${buildStdTarget}"
                 );
               RUSTFLAGS = rustflags;
+
+              preBuild = ''
+                echo "cargoVendorDir: ${crateAttrs.cargoVendorDir}"
+                echo "rustToolchain: ${cargoBuildRustToolchain'}"
+
+                # find ${crateAttrs.cargoVendorDir} -maxdepth 1 -xtype d | grep -v '^${crateAttrs.cargoVendorDir}$' | sed -E 's@(.+)@ --remap-path-prefix=\1=/@g'
```
hussein-aitlahcen, line 371: Remove this?

hussein-aitlahcen (approved):

tools/rust/crane.nix
benluelo: i kept it bc it was very useful for debugging, to see what's happening internally (especially if crane changes their vendoring logic)

PR #3906, opened 2025-03-02
Patches n stuff

benluelo:


lib/ethereum-sync-protocol/src/error.rs
```diff
@@ -93,8 +92,8 @@ pub enum Error {
     NextSyncCommitteeMismatch { expected: H384, found: H384 },
     #[error("insufficient number of sync committee participants ({0})")]
     InsufficientSyncCommitteeParticipants(usize),
-    #[error("bls error ({0:?})")]
-    Bls(AmclError),
+    // #[error("bls error ({0:?})")]
```
aeryz, line 95: let's remove this entirely if we are removing `crypto` module

lib/ethereum-sync-protocol/src/error.rs
```diff
@@ -106,9 +105,9 @@ pub enum Error {
     ClientSignatureVerification(String),
 }
 
-// NOTE: Implemented here instead of via #[from] since AmclError doesn't implement core::error::Error
-impl From<AmclError> for Error {
-    fn from(e: AmclError) -> Self {
-        Error::Bls(e)
-    }
-}
+// // NOTE: Implemented here instead of via #[from] since AmclError doesn't implement core::error::Error
```
aeryz, line 108: remove

aeryz (changes requested):

aeryz (approved):

PR #3910, opened 2025-03-03
Arbitrary light client storage

aeryz:


cosmwasm/ibc-union/core/light-client-interface/src/lib.rs
```diff
@@ -170,6 +184,50 @@ fn client_impl<T: IbcClient>(
...
+    pub client_state: Option<T::ClientState>,
+    pub consensus_state: T::ConsensusState,
+    pub storage_writes: StorageWrites,
+}
+
+/// Client creation output type
+pub struct ClientCreation<T: IbcClient> {
+    /// The client state that is going to be stored by IBC. If set to `None`, IBC will store the
+    /// client state given by the creator as is
+    pub client_state: Option<T::ClientState>,
```
benluelo, line 198: what is this used for?

cosmwasm/ibc-union/core/light-client-interface/src/lib.rs
```diff
@@ -170,6 +184,50 @@ fn client_impl<T: IbcClient>(
...
+            storage_writes: Vec::new(),
+        }
+    }
+
+    pub fn add_event(mut self, event: VerifyCreationResponseEvent) -> Self {
+        self.events.push(event);
+        self
+    }
+
+    pub fn add_storage_write(mut self, kv: (Bytes, Bytes)) -> Self {
```
benluelo, line 220: since we require `T: Decode<Self::Encoding>`, why doesn't this use `T: Encode<Self::Encoding>`?

cosmwasm/ibc-union/lightclient/ethereum/src/client.rs
```diff
@@ -96,17 +99,41 @@ impl IbcClient for EthereumLightClient {
...
         _consensus_state: &Self::ConsensusState,
-    ) -> Result<Option<Vec<VerifyCreationResponseEvent>>, IbcClientError<EthereumLightClient>> {
-        Ok(None)
+    ) -> Result<ClientCreation<Self>, IbcClientError<Self>> {
+        let Some(initial_sync_committee) = client_state.initial_sync_committee.as_ref() else {
+            return Err(Error::NoInitialSyncCommittee.into());
+        };
+        let mut client_state = client_state.clone();
+        // We only require this at the creation phase. The client then manages the committees in a separate storage.
+        client_state.initial_sync_committee = None;
```
benluelo, line 110: something we can look into perhaps is an additional "client bootstrap" state, only provided during client creation?

lib/ethereum-light-client-types/src/client_state.rs
```diff
@@ -18,6 +18,16 @@ pub struct ClientState {
     pub frozen_height: Height,
     /// the ibc contract on the counterparty chain that contains the ICS23 commitments
     pub ibc_contract_address: H160,
+    #[cfg_attr(feature = "serde", serde(default))]
+    pub initial_sync_committee: Option<InitialSyncCommittee>,
```
benluelo, line 22: add `skip_serializing_if = "Option::is_none"`

lib/ethereum-light-client-types/src/light_client_update.rs
```diff
@@ -15,11 +15,6 @@ pub enum LightClientUpdate {
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "bincode", derive(bincode::Encode, bincode::Decode))]
 pub struct EpochChangeUpdate {
```
benluelo, line 17: we don't need two separate types anymore right?

benluelo (commented):
this is a breaking change for the ethereum light client, and will require a storage migration

cosmwasm/ibc-union/core/light-client-interface/src/lib.rs
aeryz: For example in the case of Ethereum light client, `initial_sync_committee` exists under `ClientState`. But this is only to provide the sync committee at the initial state. Ethereum light client then sets this to `None` and returns the modified version of the client state so that IBC can commit that instead of committed the one that it gets from the relayer.

cosmwasm/ibc-union/core/light-client-interface/src/lib.rs
aeryz: you are correct, it would be an easier API

cosmwasm/ibc-union/lightclient/ethereum/src/client.rs
aeryz: I thought about this and initially wanted to have that, but then chose this path because it would effect the public IBC api for a specific implementation.

cosmwasm/ibc-union/core/src/contract.rs
```diff
@@ -814,6 +818,15 @@ fn create_client(
     )
 }
 
+fn store_client_data(deps: DepsMut<'_>, client_id: u32, key: Bytes, value: Bytes) {
+    deps.storage.set(
+        [client_id.to_le_bytes().as_slice(), &key]
```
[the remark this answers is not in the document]
aeryz: true, commitments might collide

cosmwasm/ibc-union/core/light-client-interface/src/lib.rs
benluelo: why not just store the original state? is the `initial_sync_committee` large?

cosmwasm/ibc-union/core/light-client-interface/src/lib.rs
```diff
@@ -108,6 +110,18 @@ impl<'a, T: IbcClient> IbcClientCtx<'a, T> {
         )
     }
 
+    pub fn read_self_storage<V: Decode<T::Encoding>>(
```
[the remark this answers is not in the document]
benluelo: i left this comment before we discussed the depolama interface, which imo is much superior to anything we can do with cw-storage-plus. also related to https://github.com/unionlabs/union/pull/3910#discussion_r1978270956

lib/ethereum-light-client-types/src/light_client_update.rs
aeryz: I'm still in favor of having the enum and such instead of optional sync committee's it makes the code messier and harder to understand.

benluelo: ah, didn't realize there were still more fields in this struct. 👍

lib/ethereum-light-client-types/src/client_state.rs
```diff
@@ -18,7 +18,7 @@ pub struct ClientState {
     pub frozen_height: Height,
     /// the ibc contract on the counterparty chain that contains the ICS23 commitments
     pub ibc_contract_address: H160,
-    #[cfg_attr(feature = "serde", serde(default))]
```
benluelo: you still need `default` lol

cosmwasm/ibc-union/core/light-client-interface/src/lib.rs
aeryz: It's total of 1024 bls pubkeys so 1024 * 384 bytes

benluelo: oof yeah lets zero that out indeed lol

benluelo: maybe leave a comment as to why we remove it

benluelo:
considering that this is a breaking change for the eth lc, lets also include the migration in this PR?

cosmwasm/ibc-union/lightclient/ethereum/src/client.rs
benluelo: ref: https://github.com/unionlabs/union/issues/3921

cosmwasm/ibc-union/core/Cargo.toml
```diff
@@ -19,10 +19,12 @@ library = []
 
 [dependencies]
 alloy                         = { workspace = true, features = ["sol-types"] }
+bincode                       = { workspace = true, features = ["alloc", "derive"], optional = true }
 cosmwasm-schema               = { workspace = true }
 cosmwasm-std                  = { workspace = true, features = ["abort"] }
 cw-storage-plus               = { workspace = true }
 ethabi                        = { workspace = true }
+ethereum-light-client-types   = { workspace = true, features = ["serde", "bincode"] }
```
benluelo, line 27: this may be resolved as i review the rest of the pr: why is this here?

cosmwasm/ibc-union/core/light-client-interface/src/lib.rs
```diff
@@ -170,6 +186,71 @@ fn client_impl<T: IbcClient>(
...
+        self.storage_writes.push((key, value.encode().into()));
+        self
+    }
+}
+
+/// Client creation output type
+pub struct ClientCreation<T: IbcClient> {
+    /// The client state that is going to be stored by IBC. If set to `None`, IBC will store the
+    /// client state given by the creator as is
+    pub client_state: Option<T::ClientState>,
```
benluelo, line 221: minor api nit: i think it would be better if the client state must always be returned as it is to be stored. removes a logic branch and makes it simpler

cosmwasm/ibc-union/core/msg/src/lightclient.rs
```diff
@@ -32,12 +24,29 @@ pub enum VerifyCreationResponseEvent {
...
+    pub events: Vec<VerifyCreationResponseEvent>,
+}
+
+#[derive(Debug, serde::Serialize, serde::Deserialize)]
+#[serde(deny_unknown_fields, rename_all = "snake_case")]
+pub struct UpdateStateResponse {
+    /// The height to save the consensus state at
+    pub height: u64,
+    /// The client state to overwrite the current one with if provided
+    pub client_state_bytes: Option<Bytes>,
```
benluelo, line 45: same comment here about always storing it

cosmwasm/ibc-union/lightclient/ethereum/src/client.rs
```diff
@@ -323,19 +345,51 @@ fn update_state<C: ChainSpec>(
...
+            sync_committee_store_key(current_epoch + 1),
+            &update.next_sync_committee,
+        );
+    }
+
+    Ok(state_update)
+}
+
+/// Additive inverse the public keys to do inverse aggregation
+fn inverse_sync_committee_pubkeys(sync_committee: &SyncCommittee) -> SyncCommittee {
```
benluelo, line 367: we may want to consider a new type for this to avoid confusion, `InverseSyncCommittee`

cosmwasm/ibc-union/lightclient/ethereum/src/verification.rs
```diff
@@ -57,41 +59,103 @@ impl BlsVerify for VerificationContext<'_> {
...
...29b2da46e2438591317258be9147359278f154e02465b938c727bb3b0c0cf8f4","0x906245e2dfb6dac3f1a7def80f2dc9eff256f0a797a8b93c742ddeb5bdddd495cb4a2c747cf0467614de98d0a636a546","0xb15460725c0d6bc3a6a7006dcf3c3e3561d9acd674c52d4199daa8598ee29eef053ae521f1271aebc66943938c9f4b7e","0xb26f5ed09f7d5bb640ec94ddd1df0b76466f69a943b4699f53d45296d5d6b8010bb61477539bc377d1a673d89074d22f","0x88e1e459ee5aeb8b36ed004f6d03da296101106fbe1b18f9bbf63e92321db51670c34050fd3b7dc56a4bad76403823ee","0x971997a5c2bbce1e8e1520da7cc84d59d6973773e541758486856856082bfba0dfc3f8ee578c69a4412b74a5fa7c808c","0xb07447c7e87459315fcbda3fb86fef27f98373b1246e2ce367e26afd87f6d698a438501fdc13cc5de9eef8d545aab768","0x887a4277ee8754733f3692a90416eeac1ebee52ff23173a827f0ba569bd84efd806eb9139049f66cc577e370d3f0962d","0xaf917d086e2e327d8d9e37ff85702536d7b15f444310d4aa832a61d850c7c3f09d31b3f5fd2a073e7fd64601275b6fca","0x838ff6630dc3908a04c51fb44a29eca5a0d88330f48c1d0dd68b8890411a394fd728f14215482b03477d33f39645dceb","0xa79ef16ee66514c2a4f34605a624dbb40d446f2637f943a4acdf0154c162fa12b30e151f03fb1a1cf100052186f94ce8"],"aggregate_pubkey":"0x8876d82074f83121a1501db0bd9a668a6fbd63a510cac73cebb36706adbf2b59a0ed547d14615a9225e51a6f2b0b6c63"}"#,
```
benluelo, line 129: nit/later work: we should prefer putting large jsons in separate files

cosmwasm/ibc-union/lightclient/ethereum/src/verification.rs
```diff
@@ -57,41 +59,103 @@ impl BlsVerify for VerificationContext<'_> {
...
+                .into_iter()
+                .map(|x| -G1Affine::deserialize_compressed(x.as_ref()).unwrap())
+                .map(|x| {
+                    let mut buf = vec![];
+                    x.serialize_compressed(&mut buf).unwrap();
+                    buf.try_into().unwrap()
+                })
+                .collect(),
+            aggregate_pubkey: next_sync_committee.aggregate_pubkey,
+        };
```
benluelo, line 144: reuse the `inverse_sync_committee_pubkeys` fn here?

lib/ethereum-light-client-types/src/client_state.rs
```diff
@@ -27,6 +27,17 @@ pub struct ClientStateV1 {
     pub frozen_height: Height,
     /// the ibc contract on the counterparty chain that contains the ICS23 commitments
     pub ibc_contract_address: H160,
+    #[cfg_attr(feature = "serde", serde(default))]
+    #[cfg_attr(feature = "serde", serde(skip_serializing_if = "Option::is_none"))]
```
benluelo, line 31: you can have multiple attrs in `cfg_attr`

cosmwasm/ibc-union/core/Cargo.toml
aeryz: yeah why is this here?

cosmwasm/ibc-union/lightclient/ethereum/src/client.rs
aeryz: I agree

cosmwasm/ibc-union/lightclient/ethereum/src/verification.rs
aeryz: yeah now that i check it in the diff view, it's huge. you can't tell when its a single line in the editor lol

aeryz: I will do this next with the tests pr btw

aeryz: #4019

cosmwasm/ibc-union/core/light-client-interface/src/lib.rs
aeryz: I initially do this to be able to do the following:
```
        Ok(ClientCreation::empty())
```
this simplifies the client codes where you absolutely dont need to care about this stuff

cosmwasm/ibc-union/core/msg/src/lightclient.rs
aeryz: This actually makes you able to save two writes when it's not needed (commitment and client state store). I kinda prefer every bit of gas optimizations whenever we can. But if you are strongly against it, we can change.

cosmwasm/ibc-union/core/Cargo.toml
aeryz: this shouldnt be here, removed

cosmwasm/ibc-union/core/msg/src/lightclient.rs
benluelo: makes sense. as long as it's well documented it should be fine

cosmwasm/ibc-union/lightclient/ethereum/Cargo.toml
```diff
@@ -23,6 +23,7 @@ ark-bls12-381                 = { version = "0.5.0" }
 ark-ec                        = { version = "0.5.0" }
 ark-serialize                 = { version = "0.5.0" }
 beacon-api-types              = { workspace = true }
+bincode                       = { workspace = true }
 bincode-1                     = { version = "1.3", package = "bincode" }
```
benluelo, line 27: do we still need bincode-1?

cosmwasm/ibc-union/core/msg/src/lightclient.rs
```diff
@@ -32,12 +24,29 @@ pub enum VerifyCreationResponseEvent {
     },
 }
 
+pub type StorageWrites = Vec<(Bytes, Bytes)>;
```
benluelo, line 27: we should use a newtype or a map structure (btreemap), some way to enforce no duplicate keys

cosmwasm/ibc-union/lightclient/ethereum/Cargo.toml
aeryz: nah removing

cosmwasm/ibc-union/core/msg/src/lightclient.rs
aeryz: I would say btreemap is the simplest

cosmwasm/ibc-union/core/light-client-interface/src/lib.rs
```diff
@@ -170,6 +186,71 @@ fn client_impl<T: IbcClient>(
...
+}
+
+impl<T: IbcClient> ClientCreation<T> {
+    pub fn empty() -> Self {
+        Self {
+            client_state: None,
+            events: Vec::new(),
+            storage_writes: Default::default(),
+        }
+    }
```
benluelo, line 236: minor nit: change this method name to `new`, and rename `ClientCreation` to `ClientCreationResult`

aeryz: promise bro, just one more nit bro

benluelo (approved):
zkgg

Issue #3915, opened 2025-03-03
Hubble cannot index non custom wasm events


0xLaurenzo:
Hubble can only index custom events from cosmwasm and is unable to index attributes emitted by a default wasm event. This is caused by [this comparison](https://github.com/unionlabs/union/blob/main/hubble/src/indexer/tendermint/block_handle.rs#L189)

```rust
const TYPE_WASM_PREFIX: &str = "wasm-";

// emitted for brevity

if !event_type.starts_with(TYPE_WASM_PREFIX)
```

This can already be solved by either changing the comparison to something like
```rust
const TYPE_WASM_PREFIX: &str = "wasm-";
const TYPE_WASM_EVENT: &str = "wasm";
// emitted for brevity

if !event_type.starts_with(TYPE_WASM_PREFIX) && !event_type == TYPE_WASM_EVENT
```

or by changing the wasm prefix like so

```rust
const TYPE_WASM_PREFIX: &str = "wasm";
// emitted for brevity

if !event_type.starts_with(TYPE_WASM_PREFIX)
```

PR #3917, opened 2025-03-03
Include wasm events in hubble indexing

0xLaurenzo:
Fixes https://github.com/unionlabs/union/issues/3915 by including events that are just "wasm" type

### DIFF
--- a/hubble/src/indexer/tendermint/block_handle.rs
+++ b/hubble/src/indexer/tendermint/block_handle.rs
@@ -156,6 +156,7 @@ fn should_include_event(
 
     // we only consider wasm events (that start with wasm-)
     const TYPE_WASM_PREFIX: &str = "wasm-";
+    const TYPE_WASM_EVENT: &str = "wasm";
 
     // property that holds the event attributes
     const ATTRIBUTES: &str = "attributes";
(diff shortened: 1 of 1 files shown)


cor:
Looks great! Can you please reword your commit to use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)? then ci will be happy :)

hubble/src/indexer/tendermint/block_handle.rs
```diff
@@ -480,6 +505,35 @@ mod tests {
         assert!(should_include);
     }
 
+    #[tokio::test]
+    async fn true_when_first_contract_address_is_active_others_are_ignored_in_a_wasm_event() {
```
qlp, line 509: Snarky: Test name doesn't match what's being tested

hubble/src/indexer/tendermint/block_handle.rs
```diff
@@ -353,6 +354,30 @@ mod tests {
         assert!(should_include);
     }
 
+    #[tokio::test]
+    async fn true_when_contract_address_is_active_in_a_wasm_event() {
```
qlp, line 358: Looks like this is a copy of the previous test?

qlp (commented):
Looking good; two snarky remarks

hubble/src/indexer/tendermint/block_handle.rs
0xLaurenzo: Updated to `true_when_first_contract_address_is_active_in_a_wasm_event_others_are_ignored`, it's a bit long and not super clear without looking at the test so could maybe still be improved

hubble/src/indexer/tendermint/block_handle.rs
0xLaurenzo: 🙈 looks like a mistaken copy paste, fixed by removing the -send_packet from the type

qlp (approved):

PR #3966, opened 2025-03-06
chore: check sync committee before verification

aeryz:


cosmwasm/ibc-union/lightclient/ethereum/src/errors.rs
```diff
@@ -55,6 +55,12 @@ pub enum Error {
 
     #[error("the misbehaviour headers must be different")]
     IdenticalMisbehaviourHeaders,
+
+    #[error("aggregate pubkey mismatch")]
+    AggregatePubkeyMismatch,
```
benluelo, line 60: why not include the keys? or will this error be removed soon

benluelo (approved):

cosmwasm/ibc-union/lightclient/ethereum/src/errors.rs
aeryz: it's gonna be removed like this week

PR #3981, opened 2025-03-07
Transfer

Swepool:


app2/src/lib/queries/channels.svelte.ts
```diff
@@ -0,0 +1,33 @@
...
+import { chains } from "$lib/stores/chains.svelte.ts"
+import { Channels } from "$lib/schema/channel.ts"
+import { channels } from "$lib/stores/channels.svelte.ts"
+
+export const channelsQuery = () =>
+  createQueryGraphql({
+    schema: Schema.Struct({ v1_ibc_union_channel_recommendations: Channels }),
+    document: graphql(`
+      query Ucs03Channels @cached(ttl: 60) {
+          v1_ibc_union_channel_recommendations(where: {_and: [{version: {_eq: "ucs03-zkgm-0"}}, {destination_chain_id: {_neq: "11155111"}}, {destination_chain_id: {_neq: "17000"}}]}) {
```
cor, line 13: you can remove the destination_chain filter, it was temporarily there for app v1

cor: also asking jurriaan if there is a `v2` api for this instead of `v1_ibc_union`

app2/src/lib/schema/channel.ts
```diff
@@ -3,3 +3,16 @@ import { Schema } from "effect"
...
+
+export class Channel extends Schema.Class<Channel>("Channel")({
+  source_port_id: Schema.String,
+  source_chain_id: Schema.String,
+  source_channel_id: Schema.Int,
+  source_connection_id: Schema.Int,
+  destination_port_id: Schema.String,
+  destination_chain_id: Schema.String,
+  destination_channel_id: Schema.Int,
+  destination_connection_id: Schema.Int
```
cor, line 15: use the ConnectoinId, ChannelId, ChainId, PortId schema types

app2/src/lib/schema/token.ts
```diff
@@ -66,3 +67,41 @@ export type Tokens = typeof Tokens.Type
...
+    type: Schema.Literal("QUOTE_LOADING")
+  }),
+  Schema.Struct({
+    type: Schema.Literal("QUOTE_MISSING_ARGUMENTS")
+  }),
+  Schema.Struct({
+    type: Schema.Literal("QUOTE_ERROR"),
+    error: Schema.String
+  })
+)
```
cor: Can do this in a next PR as this is brought over from app v1, but this can be a taggedEnum:

```typescript
export type TransferSubmission = Data.TaggedEnum<{
  Filling: {}
  SwitchChain: { state: SwitchChainState }
  ApprovalSubmit: { state: ApprovalSubmitState }
  ApprovalReceipt: { state: ApprovalReceiptState }
  TransferSubmit: { state: TransferSubmitState }
  TransferReceipt: { state: TransferReceiptState }
}>
```
https://github.com/unionlabs/union/blob/4f6eef6be9087bb1a2d8e531332ece2218846901/app2/src/lib/services/transfer-ucs03-evm/state.ts#L46

app2/src/lib/services/cosmos/rpc.ts
```diff
@@ -0,0 +1,19 @@
...
+export const cosmosRpcs: Record<CosmosChainId, string> = {
+  "elgafar-1": "https://rpc.elgafar-1.stargaze.chain.kitchen",
+  "osmo-test-5": "https://rpc.osmo-test-5.osmosis.chain.kitchen",
+  "union-testnet-9": "https://rpc.testnet-9.union.build",
+  "union-testnet-8": "https://rpc.union-testnet-8.union.chain.kitchen",
+  "stride-internal-1": "https://rpc.stride-internal-1.stride.chain.kitchen",
+  "bbn-test-5": "https://rpc.bbn-test-5.babylon.chain.kitchen"
+}
+
+export type CosmosChainId = `${(typeof cosmosChainId)[number]}`
```
cor, line 19: This should be deleted. we fetch RPCs as part of each `Chain` that should be used. check the chain store and the chain schema

app2/src/lib/services/transfer-ucs03-evm/address.ts
```diff
@@ -0,0 +1,48 @@
...
+    if (!trimmed) {
+      return yield* Effect.fail(
+        new AddressValidationError({
+          input,
+          cause: undefined
+        })
+      )
+    }
+
+    if (trimmed.toLowerCase().startsWith("0x")) {
```
cor, line 19: can use `isHex` from viem

cor (changes requested):

app2/src/lib/schema/token.ts
Swepool: created enums and saved in pr as comment, will adjust everything that is using it in next pr

app2/src/lib/services/cosmos/rpc.ts
Swepool: attacking this now then merge and start fresh with new pr.

app2/src/lib/utils/convert-display.ts
```diff
@@ -0,0 +1,137 @@
...
+/**
+ * Convert an Aptos display address (hex) to canonical bytes
+ */
+export function aptosDisplayToCanonical(displayAddress: string): Uint8Array {
+  // Validate Aptos address format (0x + 64 hex characters)
+  if (!/^0x[0-9a-fA-F]{64}$/.test(displayAddress)) {
+    throw new Error("Aptos address must be 0x followed by 64 hex characters")
+  }
+
+  // Remove 0x prefix and convert to bytes
```
cor, line 47: for these functions, make the displayAddress inputs the appropiate displayaddress types: https://github.com/unionlabs/union/blob/main/app2/src/lib/schema/address.ts

and the return type AddressCanonicalBytes. This prevents misuse (ie you put a cosmos display address into the aptos function). Once that is done, don't throw an error but return an `Option<T>`. It will in practice never be `None` unless we mess something up in our translation logic, because we have already ensured that the incoming displayAddress has the correct pattern for each of the variants as those regexes are encoded in the schema

cor (dismissed):

PR #4026, opened 2025-03-11
chore: add ethereum sync protocol tests

aeryz:


lib/ethereum-sync-protocol/src/lib.rs
```diff
@@ -386,441 +386,310 @@ pub fn next_sync_committee_gindex_at_slot<C: ChainSpec>(chain_id: u64, slot: Slo
...
+            const INVERSE: bool = false;
+
+            fn aggregate_verify_signature<'pk>(
+                &self,
+                aggregate_public_key: &'pk H384,
+                public_keys: impl IntoIterator<Item = &'pk H384>,
+                msg: Vec<u8>,
+                signature: H768,
+            ) -> Result<(), Error> {
+                let fork_version_slot =
```
benluelo, line 507: why is all of this logic done inside `aggregate_verify_signature`?

lib/ethereum-sync-protocol/src/lib.rs
```diff
@@ -386,441 +386,310 @@ pub fn next_sync_committee_gindex_at_slot<C: ChainSpec>(chain_id: u64, slot: Slo
...
+    }
+
+    const SEPOLIA_CHAIN_ID: u64 = 11155111;
+
+    lazy_static! {
+        pub static ref UPDATE_6553725: ethereum_sync_protocol_types::LightClientUpdate =
+            serde_json::from_str(&include_str!("./test/light_client_update_6553725.json")).unwrap();
+        pub static ref SYNC_COMMITTEE_6553725: SyncCommittee =
+            serde_json::from_str(include_str!("./test/sync_committee_6553725.json")).unwrap();
+    }
```
benluelo, line 420: why not use https://doc.rust-lang.org/stable/std/cell/struct.LazyCell.html

lib/ethereum-sync-protocol/src/lib.rs
```diff
@@ -386,441 +386,310 @@ pub fn next_sync_committee_gindex_at_slot<C: ChainSpec>(chain_id: u64, slot: Slo
...
+
+    #[test]
+    fn validate_update_works() {
+        assert_eq!(update_6553725(&UPDATE_6553725), Ok(()));
+    }
+
+    #[test]
+    fn validate_update_fails_when_sync_committee_is_wrong() {
+        let mut update = UPDATE_6553725.clone();
+        update.sync_aggregate.sync_committee_bits = Vec::new();
```
benluelo, line 449: this should technically fail, since the vector length is incorrect, so:

- add a test that ensures the sync committee bits vector length is checked correctly
- use https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.clear here instead

hussein-aitlahcen:
please add ci checks for all crates where you introduced tests, we no longer run workspace tests in the ci (they must be explicity defined)

benluelo (commented):
general comment: you use the pattern `<something>_height` a lot, imo it would be more readable to have something like `mod data_<height> { const <something>: ... }`

lib/ethereum-sync-protocol/src/lib.rs
aeryz: this test doesn't do signature verification since it isn't the library's responsibility to contain the implementation of the signature verification. This checks whether the provided data to signature verification is valid for a certain scenario.

lib/ethereum-sync-protocol/src/lib.rs
```diff
@@ -386,441 +386,310 @@ pub fn next_sync_committee_gindex_at_slot<C: ChainSpec>(chain_id: u64, slot: Slo
...
-// }
+#[cfg(test)]
+mod tests {
+    use beacon_api_types::chain_spec::Mainnet;
+    use hex_literal::hex;
+    use lazy_static::lazy_static;
+
+    use super::*;
+
+    pub struct BlsVerifier;
```
benluelo, line 397: rename to `AlwaysSuccessBlsVerifier` to prevent confusion

lib/ethereum-sync-protocol/src/lib.rs
```diff
@@ -386,441 +386,310 @@ pub fn next_sync_committee_gindex_at_slot<C: ChainSpec>(chain_id: u64, slot: Slo
...
+            update_6553725(&update),
+            Err(Error::InvalidMerkleBranch(..))
+        ));
+    }
+
+    /// This test also ensures that any `attested_header.beacon` alteration would cause the signature verification
+    /// to fail.
+    #[test]
+    fn validate_update_signature_data_correct() {
+        pub struct BlsVerifier;
```
benluelo, line 495: rename to `AssertDataCorrectBlsVerifier` to prevent confusion

benluelo (approved):

benluelo (approved):

PR #4036, opened 2025-03-12
fix(zkgm): revert batch if any instruction require a market maker

hussein-aitlahcen:
Also add some assertions on the batch acknowledgements to make sure upgrades don't break the semantic.


aeryz:
Tagging @Caglankaan for reflecting this stuff to move

aeryz (dismissed):

PR #4043, opened 2025-03-13
feat(ibc-union): alpha omega

hussein-aitlahcen:
- simplify commitment for packets/acks
  - remove channel_id prefix
  - batch_send/batch_acks ensures packets are batched for same channel
- fix forward ack/timeout incorrectly checking packet hash instead of salt
- prefix cosmwasm store and forbids direct call to deps.storage.get, prefer read_commit
- refactor voyager for the new interface


cosmwasm/ibc-union/core/src/contract.rs
```diff
@@ -1677,24 +1596,23 @@ fn send_packet(
...
 
     Ok(Response::new()
-        .add_event(Event::new(events::packet::SEND).add_attribute(
-            events::attribute::PACKET,
-            serde_json::to_string(&packet).expect("packet serialization is infallible; qed;"),
-        ))
-        .set_data(to_json_binary(&packet)?))
+        .add_event(
+            Event::new(events::packet::SEND)
+                .add_attribute(events::attribute::PACKET, &serialized_packet),
```
benluelo, line 1613: can we include the packet fields as individual attributes?

cosmwasm/ibc-union/core/src/contract.rs
```diff
@@ -103,6 +110,39 @@ pub mod events {
...
+        (
+            events::attribute::PACKET_TIMEOUT_HEIGHT,
+            packet.timeout_height.to_string(),
+        ),
+        (
+            events::attribute::PACKET_TIMEOUT_TIMESTAMP,
+            packet.timeout_timestamp.to_string(),
+        ),
+    ]
+    .map(Into::into)
```
benluelo, line 134: one thing to note: `array::map` is [quite inefficient](https://doc.rust-lang.org/stable/std/primitive.array.html#note-on-performance-and-stack-usage), so we may want to consider returning `impl Iterator<Item = Attribute>` here instead

evm/contracts/core/04-channel/IBCChannel.sol
```diff
@@ -11,43 +11,43 @@ import "../../lib/Hex.sol";
 
 library IBCChannelLib {
     event ChannelOpenInit(
-        address portId,
-        uint32 channelId,
+        address indexed portId,
+        uint32 indexed channelId,
         bytes counterpartyPortId,
         uint32 connectionId,
         string version
```
benluelo, line 18: maybe we can index version as well? useful for protocol developers

cosmwasm/ibc-union/core/src/contract.rs
hussein-aitlahcen: Seems to be fine for small array?
> This mainly concerns large arrays, as mapping over small arrays seem to be optimized just fine. 
Also this is a constant sized array so it should be completely erased and inlined I assume

benluelo: see: https://godbolt.org/z/6hxrh4rYx

benluelo: (wasm: https://godbolt.org/z/Y1sMaxjna)

benluelo (approved):
neat

PR #4047, opened 2025-03-13
Movement fe integration appv2

Caglankaan:


app2/src/lib/components/Transfer/index.svelte
```diff
@@ -63,6 +76,8 @@ function getError(state: TransferStateUnion): string | null {
       }
       return null
     }
+    default:
+      return null
```
Swepool, line 80: We return nulls here, but thats ok for now as this will be refactored to fit new design

app2/src/lib/services/aptos/clients.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export const getPublicClient = (chain: Chain) =>
+  Effect.gen(function* () {
+    if (chain.rpc_type !== "aptos") {
+      throw new NoAptosChainError({ chain })
+    }
+    const aptosClient = yield* Effect.try({
+      try: () => {
+        const config = new AptosConfig({
+          fullnode: "https://aptos.testnet.bardock.movementlabs.xyz/v1",
```
Swepool, line 33: I think here we should get rpc url from chain

`chain.getRpcUrl("rpc")`

app2/src/lib/services/aptos/clients.ts
```diff
@@ -0,0 +1,62 @@
...
+export const getWalletClient = (chain: Chain) =>
+  Effect.gen(function* () {
+    if (chain.rpc_type !== "aptos") {
+      throw new NoAptosChainError({ chain })
+    }
+
+    const aptosClient = yield* Effect.try({
+      try: () => {
+        const config = new AptosConfig({
+          fullnode: "https://aptos.testnet.bardock.movementlabs.xyz/v1",
```
Swepool, line 52: maybe same here

`chain.getRpcUrl("rpc")`

app2/src/lib/services/transfer-ucs03-aptos/channel.ts
Swepool, line 1: I think we can move channel into shared folder and use for evm, cosmos, aptos. maybe in another pr tho

app2/src/lib/services/transfer-ucs03-aptos/quote-token.ts
Swepool, line 1: You think we can use the fn in shared folder?

app2/src/lib/services/transfer-ucs03-aptos/salt.ts
Swepool, line 1: we now have salt in shared folder

Swepool (changes requested):

app2/src/lib/services/transfer-ucs03-aptos/salt.ts
Caglankaan: that 2 salt is different, added that salt in to the shared folder

cor (approved):

Swepool (approved):

PR #4050, opened 2025-03-13
feat(hubble): testnet feature for lenient rpc handling

qlp:


hubble/hubble.nix
```diff
@@ -101,6 +101,11 @@
                 example = 1;
                 default = 200;
               };
+              options.testnet = mkOption {
+                type = types.nullOr types.bool;
+                default = null;
+                description = "Testnet (default false)";
```
benluelo, line 107: default = null (default false)

benluelo (approved):

hubble/hubble.nix
qlp: default value in produced json is null. hubbles default is false.

reason is that I don't want a big json file and let hubble control what's the default

Chinu143blue:
Super nice project

cornelboy:
Keep building 🔥

PR #4063, opened 2025-03-17
Sequential rank-ordered balance fetcher

cor:
- **feat(app2): rank in token schema**
- **feat(app2): sequential token fetcher**
- **refactor(app2): more effective balance fetcher**
- **feat(app2): query tokens by rank**
- **feat(app2): general token blacklist mechanism**
- **fix(app2): cleanup**


Swepool (approved):
looks solid, like the queue, will check their docs.

PR #4072, opened 2025-03-18
Transfer improvements

Swepool:


app2/src/lib/components/Transfer/validation.ts
```diff
@@ -0,0 +1,88 @@
...
+  type EVMTransfer,
+  TransferSchema
+} from "$lib/schema/transfer-args.ts"
+import type { TransferSubmission as EvmTransferSubmission } from "$lib/services/transfer-ucs03-evm"
+import type { TransferSubmission as CosmosTransferSubmission } from "$lib/services/transfer-ucs03-cosmos"
+import type { TransferSubmission as AptosTransferSubmission } from "$lib/services/transfer-ucs03-aptos"
+
+export type TransferState = Data.TaggedEnum<{
+  Empty: {}
+  EVM: { state: EvmTransferSubmission }
```
cor, line 14: should be Evm not EVM

app2/src/lib/schema/chain.ts
```diff
@@ -39,17 +39,17 @@ export type RpcProtocolType = typeof RpcProtocolType.Type
 
 export class Rpc extends Schema.Class<Rpc>("Rpc")({
   type: RpcProtocolType,
-  url: Schema.URL
```
cor, line 42: why? it is an url right?

app2/src/lib/services/shared/quote-token.ts
```diff
@@ -60,11 +67,14 @@ export const getQuoteToken = (
             address: channel.destination_port_id,
             abi: ucs03ZkgmAbi,
             functionName: "predictWrappedToken",
-            args: [0, channel.destination_channel_id, base_token]
+            args: [BigInt(0), channel.destination_channel_id, base_token]
```
cor, line 70: i think you can just do 0n here

app2/src/lib/services/transfer-ucs03-evm/chain.ts
```diff
@@ -1,15 +1,25 @@
-import { Effect } from "effect"
+import { Effect, Option } from "effect"
 import { switchChain as wagmiSwitchChain } from "@wagmi/core"
 import { wagmiConfig, type ConfiguredChainId } from "$lib/wallet/evm/wagmi-config"
 import { SwitchChainError } from "./errors.ts"
 import type { SwitchChainErrorType } from "viem"
+import type { Chain } from "$lib/schema/chain.ts"
 
-export const switchChain = (chainId: ConfiguredChainId) =>
+export const switchChain = (chain: Chain) =>
```
cor, line 8: better to make this take a Viem Chain instead and do the conversion outside of this function

cor (changes requested):
Looking good! few minor comments

app2/src/lib/services/transfer-ucs03-evm/chain.ts
Swepool: Fixed

app2/src/lib/schema/chain.ts
cor: super weird, idk either. we'll keep string for now then 👍

cor (approved):

cor (approved):

PR #4102, opened 2025-03-21
frissíthető

benluelo:
- unionlabs-cosmwasm-upgradeable -> frissitheto
- massive overhaul to our crane builders
  - no more individual crate checks by default, the entire workspace is now clippy'd and tested at once
  - remove flake modules that only export individual crate tests
  - all builds now export a .release as well as their package (`nix build .#ibc-union`, `nix build .#ibc-union.release`)
- add `embed-commit` library, enabling verifiable release builds of any rust artifact by embedding the commit hash at build time
  - add `embed-commit-verifier` that extracts the commit from the built artifact
- other stuff i probably forgot


hussein-aitlahcen (approved):

lib/embed-commit/verifier/src/main.rs
```diff
@@ -0,0 +1,90 @@
...
+}
+
+fn main() -> Result<()> {
+    match App::parse() {
+        App::Extract { path } => {
+            let file = std::fs::read(path).context("reading input artifact")?;
+
+            let rev = match &file.get(0..4) {
+                Some(b"\0asm") => extract_wasm(&file)?,
+                Some(b"\0elf") => extract_elf(&file)?,
```
aeryz, line 24: nit: use `|`

aeryz (approved):

lib/embed-commit/verifier/src/main.rs
benluelo: I actually think the first elf branch should be removed, I'm not sure where I found those magic bytes. will test tomorrow

PR #4107, opened 2025-03-24
refactor: remove deprecated extract_if usage 

Himess:
This PR addresses compilation errors that occur due to the recent stabilization of Vec::extract_if and removes the now-invalid nightly feature flag:

Removed `#![feature(extract_if)]` as `extract_if` is stable since Rust 1.87.0.

`extract_if`  Error:

Pasted Graphic 16

Replaced `Vec::extract_if` usage with manual `.retain()` and matching logic to maintain compatibility with both stable and nightly toolchains.

AFTER
`cargo +nightly check -p unionlabs` 

Pasted Graphic 19

⚠️ Some sqlx-related errors still exist due to missing DB env, but unrelated to this PR.

### DIFF
--- a/lib/cosmos-sdk-event/src/lib.rs
+++ b/lib/cosmos-sdk-event/src/lib.rs
@@ -1,5 +1,3 @@
-#![feature(extract_if)]
-
 use std::{collections::BTreeMap, fmt::Display, str::FromStr};
 
 use cometbft_types::abci::event_attribute::EventAttribute;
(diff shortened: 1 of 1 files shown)


benluelo:
thanks for the PR! since we pin a rust nightly version in our repo (see [here](https://github.com/unionlabs/union/blob/0d9fe9a1443c90cbe868f3cdd33f1046ede515d8/tools/rust/rust.nix#L13)), we won't do this for now. i will however be bumping our nightly version very soon, which will resolve this. in general, when working either in our repo or using our libraries externally, we highly recommend using the exact pinned rust version we use.

Himess:
Thanks for the reply @benluelo . I'm trying to contribute as much as I can. I'll make sure to follow what you mentioned.

PR #4131, opened 2025-03-28
feat(deployment): add channels_experimental and universal_chain_id

cor:


deployments/deployments-testnet-10.json
```diff
@@ -120,9 +126,11 @@
         }
       }
     },
-    "channels": [1]
+    "channels": [1],
+    "channels_experimental": []
```
benluelo, line 130: can be omitted if empty

cor: fixed here: a551788e2c7d

benluelo (approved):

benluelo (approved):

qlp (approved):

PR #4137, opened 2025-03-30
[Proposal] Use TypeScript Project References

ehegnes:
Given colocation of package and app sources, this repository lends itself well to engendering further mono-repository tooling.

This PR proposes utilizing [TypeScript Project References](https://www.typescriptlang.org/docs/handbook/project-references.html) to source inter-repository packages locally from the repository rather than relying on published artifacts.

**Implications:**
- No longer must packages be published publicly before utilizing package changes in consuming applications.
- All workspace enabled projects in the repository depend directly on the live code for colocated packages, rather than on published artifacts.
- Package changes can now be "dogfooded" prior to planned public release.
- LSP developer experience is improved such that "go-to" calls point directly to in-repository source rather than `node_modules/*`.
- `test/*` files can now import code exactly as an artifact consumer would (e.g. `import * as _ from "@unionlabs/sdk"` rather than `import * as _ from "../src/index.js"`) which allows for unit tests to serve as usage examples.

It is proposed to leverage PNPM over NPM to additionally improve mono-repository tooling:
- PNPM should be preferred over NPM for the following reasons:
  - Mature handling for workspaces.
  - Defense against [phantom dependencies](https://rushjs.io/pages/advanced/phantom_deps/).
  - Defense against [NPM doppelgangers](https://rushjs.io/pages/advanced/npm_doppelgangers/).
  - Strict handling for peer dependencies.
- PNPM workspaces allows for patches to be declared once and inherited by all workspace projects (without the need for `patch-package`).

This PR additionally introduces, for `@unionlabs/sdk`, the same build system as `effect` which produces tree-shakeable module-based path imports in both CJS and ESM formatting.

**Miscellaneous:**
- Some packages (e.g. `effect` and `viem`) have been moved to `peerDependencies` where appropriate and gated by major version number such that consumers are not at risk of bundling multiple versions.
- `"DOM"` added to `@unionlabs/client` TS config `"lib"` entry to ensure global `Crypto` API is well-defined, which was previously pointing to a `bun` stub.
- [`dpdm`](https://github.com/acrazing/dpdm) added to `@unionlabs/sdk` to detect circular dependencies.
- Path-based imports (e.g. `#/path/after/package/src`) have been removed in `@unionlabs/client` as it provides no benefit and causes issues with project references.


nix/buildPnpmPackage.nix
```diff
@@ -0,0 +1,61 @@
+# buildPnpmPackage.nix
```
ehegnes, line 1: To make building PNPM packages easier.

@cor note the [`fileset`](https://nixos.org/manual/nixpkgs/stable/#sec-functions-library-fileset) usage which constructs a source derivation from only required files.

PR #4152, opened 2025-04-02
chore(deployment): separate channels

qlp:


deployments/channels-testnet-10.json
```diff
@@ -0,0 +1,27 @@
+{
+  "_note-1": "only register the 'init' side of the channel",
+  "_note-2": "keep universal chain ids (string) and channel (number) in ascending order",
+  "_note-3": "keep comments consistent, but don't parse them",
```
benluelo, line 4: are these ingested by any tooling? if not, this file could be jsonc?

benluelo (commented):
let's also remove the channels information from the deployments*.json files

qlp:
> let's also remove the channels information from the deployments*.json files

@benluelo done (didn't want to do it to prevent conflicts on your end)

benluelo (approved):

PR #4158, opened 2025-04-03
Clients json

benluelo:


deployments/clients.jsonc
```diff
@@ -5,12 +5,18 @@
     "10": "bbn-test-5"
   },
   "bbn-test-5": {
-    "5": "union-testnet-10"
+    "5": "union-testnet-10",
```
qlp, line 8: use universal chain ids?

(just to make it a habit)

cor: agree it should be

benluelo: I consume this in Voyager infra so for now no

PR #4164, opened 2025-04-04
feat(app2): improve balance polling and supervision

ehegnes:


ehegnes:
Superseded by a separate change.

PR #4175, opened 2025-04-05
chore: replace `futures::mpsc` with `tokio::mpsc` for Axum compatibility

mdqst:
I replaced `futures::mpsc::UnboundedSender` with `tokio::sync::mpsc::UnboundedSender` in the Axum server setup. The original sender doesn't implement `Sync`, which causes issues when used with `axum::extract::State`.  

`tokio::mpsc::UnboundedSender` implements `Clone + Send + Sync + 'static`, making it a better fit here and removing the need to wrap the sender in `Arc<Mutex<...>>`. The fix makes the enqueue route work as intended without extra boilerplate.

### DIFF
--- a/voyager/src/api.rs
+++ b/voyager/src/api.rs
@@ -3,11 +3,7 @@ use std::net::SocketAddr;
 use axum::{
     extract::State,
     routing::{get, post},
-    Json,
-};
-use futures::{
-    channel::mpsc::{unbounded, UnboundedReceiver, UnboundedSender},
-    SinkExt,
+    Json, Router,
 };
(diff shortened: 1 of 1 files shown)


benluelo:
thanks for the PR! was there an issue with the current design that this PR fixes? you mention `Arc<Mutex<...>>`, but we don't use that anywhere in the router.

benluelo:
I'm still not sure what problem you're trying to solve? The code works as it is right now.

benluelo:
closing this as it does not seem to solve anything.

PR #4176, opened 2025-04-05
fix(ibc-union): ensure batches share the same channel

hussein-aitlahcen:
- ensure that the batched packets share the same channel, avoiding a malicious client from accepting a batch that would result in arbitrary packets indexed at i+1 to be accepted by the protocol


benluelo (approved):
can we add a test for this?

benluelo (approved):

PR #4202, opened 2025-04-08
feat(trusted-mpt): add pubkey auth

aeryz:


lib/trusted-mpt-light-client-types/src/signed_data.rs
```diff
@@ -0,0 +1,30 @@
+use unionlabs::{
+    encoding::{Bincode, Encode},
+    primitives::{H256, H512},
+};
+
+#[derive(Debug, Clone, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "bincode", derive(bincode::Encode, bincode::Decode))]
+pub struct SignedData<T: Encode<Bincode>> {
+    data: T,
```
benluelo, line 10: fields should be pub since you derive serde

lib/trusted-mpt-light-client-types/src/signed_data.rs
```diff
@@ -0,0 +1,30 @@
+use unionlabs::{
+    encoding::{Bincode, Encode},
+    primitives::{H256, H512},
+};
+
+#[derive(Debug, Clone, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "bincode", derive(bincode::Encode, bincode::Decode))]
+pub struct SignedData<T: Encode<Bincode>> {
```
benluelo, line 9: can't use bincode as it's feature gated, just use raw T here

lib/trusted-mpt-light-client-types/src/signed_data.rs
```diff
@@ -0,0 +1,30 @@
...
+    data: T,
+    signature: H512,
+}
+
+impl<T: Encode<Bincode> + Clone> SignedData<T> {
+    pub fn new(data: T, signature: H512) -> Self {
```
benluelo, line 15: remove constructor since fields are (now) public

lib/trusted-mpt-light-client-types/src/signed_data.rs
```diff
@@ -0,0 +1,30 @@
...
+        SignedData { data, signature }
+    }
+
+    pub fn unwrap_verified<V: Fn(&[u8], H512, H256) -> Option<bool>>(
```
benluelo, line 19: put the Encoding param on this function (E: Encoding, and then add where T: Encode<E>)

lib/trusted-mpt-light-client-types/src/signed_data.rs
aeryz: we talked through this and decided to keep this private and document the reason

lib/trusted-mpt-light-client-types/src/signed_data.rs
```diff
@@ -0,0 +1,45 @@
...
+        E: Encoding,
+        T: Encode<E>,
+    {
+        SignedData {
+            signature: sign(&data.clone().encode()),
+            data,
+        }
+    }
+
+    pub fn unwrap_verified<E, V: Fn(&[u8], H512, H256) -> Option<bool>>(
```
benluelo, line 28: nit: either move the E bound to the params list or all bounds to the where clause

lib/trusted-mpt-light-client-types/src/signed_data.rs
```diff
@@ -0,0 +1,45 @@
...
+    encoding::{Encode, Encoding},
+    primitives::{H256, H512},
+};
+
+#[derive(Debug, Clone, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "bincode", derive(bincode::Encode, bincode::Decode))]
+pub struct SignedData<T> {
+    data: T,
+    signature: H512,
```
benluelo, line 11: another small nit, let's make an issue for it to be done later: this type could be generic over the signature size as well. we could then look into moving this type into somewhere more generic. for now this is fine though

lib/trusted-mpt-light-client-types/src/signed_data.rs
```diff
@@ -0,0 +1,45 @@
...
+}
+
+pub struct SignatureVerificationFailure;
```
benluelo, line 14: impl Error for this (can derive thiserror::Error and use the #[error()] attribute)

lib/trusted-mpt-light-client-types/src/signed_data.rs
benluelo: https://github.com/unionlabs/union/issues/4207

benluelo (approved):

PR #4226, opened 2025-04-09
feat(cosmwasm): ibc-union/zkgm access control

hussein-aitlahcen:
- add rate limit operators on zkgm
- add whitelist based relayers on ibc-union

### DIFF
--- a/cosmwasm/ibc-union/core/src/contract.rs
+++ b/cosmwasm/ibc-union/core/src/contract.rs
@@ -5,7 +5,7 @@ use alloy::sol_types::SolValue;
 use cosmwasm_std::entry_point;
 use cosmwasm_std::{
     to_json_binary, wasm_execute, Addr, Attribute, Binary, Deps, DepsMut, Env, Event, MessageInfo,
-    OverflowError, OverflowOperation, Response, StdError, StdResult,
+    OverflowError, OverflowOperation, Response, StdError, StdResult, Storage,
 };
 use depolama::{RawStore, StorageExt};
 use frissitheto::{UpgradeError, UpgradeMsg};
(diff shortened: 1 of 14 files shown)


cosmwasm/ibc-union/core/src/contract.rs
```diff
@@ -431,6 +462,26 @@ pub fn execute(
...
+        ExecuteMsg::RemoveRelayer(relayer) => {
+            ensure_relayer_admin(deps.storage, &info.sender)?;
+            let relayer = deps.api.addr_validate(&relayer)?;
+            deps.storage.write::<WhitelistedRelayers>(&relayer, &());
+            Ok(Response::new().add_event(
+                Event::new("whitelisted_relayers")
+                    .add_attribute("action", "revoke")
+                    .add_attribute("relayer", relayer),
+            ))
+        }
```
benluelo, line 484: backwards delete/write

cosmwasm/ibc-union/core/src/state.rs
```diff
@@ -384,3 +384,42 @@ fn read_fixed_bytes<const N: usize>(raw: &Bytes) -> StdResult<[u8; N]> {
...
+
+impl ValueCodec<()> for WhitelistedRelayers {
+    fn encode_value(_: &()) -> Bytes {
+        Bytes::default()
+    }
+
+    fn decode_value(_: &Bytes) -> StdResult<()> {
+        Ok(())
+    }
+}
```
benluelo, line 425: im not sure if using an empty vec is safe, empty/missing storage in cosmwasm is a bit whack. we can just write one byte maybe?

benluelo: i would also actually decode the key for sanity check (just ensure it's `[0x00]`)

benluelo (approved):

benluelo (dismissed):

benluelo (approved):

PR #4243, opened 2025-04-10
app2 unwrap

cor:
- **fix(app2): whitelist sepolia**
- **feat(ts-sdk): token wrapping query**
- **fix(ts-sdk): quote token stuff**
- **fix(ts-sdk): invert source/dest channel on unwrap**
- **fix(app2): unwrap transfer**
- **refactor(ts-sdk): always transform graphql denoms**


Swepool (approved):
I need this to continue, please open up another pr if we need to change something.

PR #4271, opened 2025-04-11
feat(app2): add receiver check step

ehegnes:


Swepool (approved):
lgtm we ran a few tests as well

PR #4336, opened 2025-04-14
feat(hubble): subscribe to github files

qlp:


hubble/src/github_fetcher/fetcher.rs
```diff
@@ -0,0 +1,142 @@
...
+use crate::github_fetcher::{
+    self,
+    client::{fetch_commit_details, fetch_file_contents, CommitDetails, FileDownloadError},
+    postgres::get_subscriptions,
+    Attempt, Download, Subscription,
+};
+
+#[derive(Debug, thiserror::Error)]
+pub enum UpdateSubscriptionError {
+    #[error("database error creation transaction for subscription {0}: {1}")]
```
benluelo, line 16: minor nit, non blocking: prefer to use `#[source]` instead of displaying the error (`sqlx::Error` in this case) as you will lose all sources other than the top level error

benluelo (approved):

PR #4343, opened 2025-04-14
feat(ibc-union): deprecated height timeout

hussein-aitlahcen:
- deprecate and block the usage of timeout height

### DIFF
--- a/cosmwasm/ibc-union/core/src/contract.rs
+++ b/cosmwasm/ibc-union/core/src/contract.rs
@@ -1769,15 +1769,12 @@ fn process_receive(
             return Err(ContractError::BatchSameChannelOnly);
         }
 
-        if packet.timeout_height > 0 && (env.block.height >= packet.timeout_height) {
-            return Err(ContractError::ReceivedTimedOutPacketHeight {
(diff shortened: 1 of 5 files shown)


cosmwasm/ibc-union/core/src/contract.rs
```diff
@@ -1897,7 +1894,11 @@ fn send_packet(
     timeout_timestamp: Timestamp,
     data: Vec<u8>,
 ) -> ContractResult {
-    if timeout_timestamp.is_zero() && timeout_height == 0 {
+    if timeout_height != 0 {
+        return Err(ContractError::TimeoutHeightUnsupported);
+    }
```
benluelo, line 1899: would be good to have a test for this such that we don't accidentally accept non-zero heights

benluelo (approved):

cosmwasm/ibc-union/core/src/contract.rs
hussein-aitlahcen: this is just informational the semantic has been removed from the receiving/timeout side so it's effectively a noop even if it passes

PR #4348, opened 2025-04-15
chore: update deployment

hussein-aitlahcen:


deployments/deployments.json
```diff
@@ -242,8 +242,8 @@
       "sender": "0x95fb5cb304508d74d855514d7bc9bda75c304ce2",
       "core": {
         "address": "0xee4ea8d358473f0fcebf0329feed95d56e8c04d7",
-        "height": 22242649,
-        "commit": "33462cd11c9fc68deab4c1d6db3dada33e1259ef"
+        "height": 22273483,
```
benluelo, line 245: heights shouldn't update, we need to fix the script I guess?

benluelo (changes requested):

deployments/deployments.json
hussein-aitlahcen: oof, it's already been updated a couple of time I believe

hussein-aitlahcen: got it, we fetch the initialized height of the impl instead of proxy

benluelo (approved):

benluelo (dismissed):

benluelo (approved):

PR #4356, opened 2025-04-15
chore: add trusted mpt consensus

aeryz:


lib/voyager-primitives/src/lib.rs
```diff
@@ -200,6 +200,9 @@ impl ConsensusType {
     /// [Bob]: https://github.com/ethereum-optimism/optimism/blob/v1.7.2/packages/contracts-bedrock/src/L1/L2OutputOracle.sol
     pub const BOB: &'static str = "bob";
 
+    /// EVM chain consensus that will be tracked in a trusted way. Development purposes only.
+    pub const TRUSTED_MPT: &'static str = "trusted/evm/mpt";
```
benluelo, line 204: this is technically just trusted/evm, since the "consensus" is just the eth_getBlockNumber response (doesn't rely on any consensus layer)

aeryz: correct

lib/voyager-primitives/src/lib.rs
```diff
@@ -200,6 +200,15 @@ impl ConsensusType {
...
     pub const BOB: &'static str = "bob";
 
+    /// Trusted consensus of an [Ethereum JSON-RPC] compatible chain.
+    ///
+    /// <div class="warning">
+    /// Blocks are considered finalized after a configurable delay period. This is not secure, and should only be used in development or testing environments.
+    /// </div>
+    ///
+    /// [Ethereum JSON-RPC]: https://ethereum.github.io/execution-apis/api-documentation/
+    pub const TRUSTED_MPT: &'static str = "trusted/evm";
```
benluelo, line 210: rename to `TRUSTED_EVM`

PR #4368, opened 2025-04-17
Use a Timestamp and Duration newtype throughout the stack

benluelo:
no more confusion about whether something is nanos or secs


cosmwasm/ibc-union/lightclient/ethereum/src/client.rs
```diff
@@ -597,10 +597,10 @@ mod tests {
             slot: INITIAL_HEADER.slot,
             state_root: INITIAL_HEADER.state_root,
             storage_root: INITIAL_STORAGE_HASH,
-            timestamp: compute_timestamp_at_slot::<Mainnet>(
+            timestamp: Timestamp::from_secs(compute_timestamp_at_slot::<Mainnet>(
                 SEPOLIA_GENESIS_TIME,
                 INITIAL_HEADER.slot,
-            ),
+            )),
```
benluelo, line 603: @aeryz can you confirm this is the intended behaviour?

aeryz (approved):
A very much needed PR, gj

cosmwasm/ibc-union/lightclient/ethereum/src/client.rs
aeryz: Onayliyorum

benluelo: discussed internally and realized that while the tests were "incorrect" before (storing the timestamp as seconds instead of nanoseconds), it didn't cause any issues since the timestamp is only read when doing packet verification.

aeryz (approved):

PoisonPhang (approved):

PR #4369, opened 2025-04-17
Hubble abi fetcher

qlp:
generate abis when they're required for processing evm events


hubble/src/abi_fetcher/fetcher.rs
```diff
@@ -0,0 +1,117 @@
...
+        postgres::{get_missing_abi_dependencies, insert_download},
+        AbiDependency, Attempt, Download,
+    },
+    github_client::commit_details::CommitDetailsError,
+};
+
+#[derive(Debug, thiserror::Error)]
+pub enum FetchAbiError {
+    #[error("database error creation transaction for dependency {0}: {1}")]
+    CreateTransaction(AbiDependency, #[source] sqlx::Error),
```
benluelo, line 19: don't print an error in the display if it's also a source, otherwise you'll get duplicated messages when printing the full trace

hubble/src/abi_fetcher/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+
+use serde_json::Value;
+
+mod client;
+mod fetcher;
+mod postgres;
+
+#[derive(Clone, Debug)]
+pub struct AbiDependency {
+    commit: Vec<u8>,
```
benluelo, line 11: a commit is a fixed 20 byte hash

hubble/src/abi_fetcher/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+            hex::encode(&self.commit),
+            self.success,
+            self.details
+        ))
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct Download<'a> {
+    commit: &'a Vec<u8>,
```
benluelo, line 40: this should be the same `[u8; 20]` as i mentioned above, but in general you almost never want `&Vec<T>`, `&[T]` is preferred

hubble/src/abi_fetcher/client.rs
```diff
@@ -0,0 +1,182 @@
...
+        .arg(nix_build_argument)
+        .current_dir(&work_dir)
+        .output()
+        .map_err(BuildError::CommandError)?;
+
+    info!("abi build: {} => {}", nix_command, output.status);
+
+    let mut log = String::new();
+    log.push_str(&String::from_utf8_lossy(&output.stdout));
+    log.push_str(&String::from_utf8_lossy(&output.stderr));
```
benluelo, line 102: nit: you can use `+=`

hubble/src/abi_fetcher/client.rs
```diff
@@ -0,0 +1,182 @@
...
+
+    debug!("abi build log: \n-----------\n{}\n-----------", log);
```
benluelo, line 104: nit: in general, tracing logs should be single-line and not contain newlines. i would recommend putting the stdout and stderr as log fields (i.e. something like `debug!(%output.stdout, %output.stderr, "abi build log", log);`)

hubble/src/abi_fetcher/fetcher.rs
qlp: but you won't see it if you're not printing the trace?

hubble/src/abi_fetcher/client.rs
qlp: decided to split in stdout/stderr

hubble/src/abi_fetcher/client.rs
qlp: 👍

hubble/src/abi_fetcher/mod.rs
qlp: decided to create a new type

hubble/src/abi_fetcher/mod.rs
qlp: not the first time you mentioned; thanks for you patience 😄

hubble/src/abi_fetcher/fetcher.rs
benluelo: you either need to set the source or print it in display, but not both. i have [this type](https://github.com/unionlabs/union/blob/752d7221d5725012a215e6795f5d89957212d257/lib/unionlabs/src/lib.rs#L243) that can be used to print the entire error trace in display.

benluelo (approved):

PR #4393, opened 2025-04-19
feat(docs): leap kelpr denom add

ehegnes:
Credit @PoisonPhang.
Closes #4389.


ehegnes:
Already resolved by another change.

PR #4398, opened 2025-04-21
feat(zkgm): optional rate limiting

hussein-aitlahcen:


PoisonPhang (dismissed):

cosmwasm/ibc-union/app/ucs03-zkgm/src/contract.rs
```diff
@@ -2052,6 +2053,10 @@ pub fn migrate(
             Ok((res, None))
         },
         |deps, migrate_msg, _current_version| {
+            CONFIG.update::<_, ContractError>(deps.storage, |mut config| {
+                config.rate_limit_disabled = migrate_msg.rate_limit_disabled;
+                Ok(config)
+            })?;
```
benluelo, line 2059: small nit that can be done later, but we should document the semantics of all of the MigrateMsg fields, and how additional parameters should be added in the future

benluelo (approved):
lgtm. our deployment scripts are getting pretty unwieldy though, we should work on a better design for those soon

benluelo (approved):

benluelo (approved):

PR #4402, opened 2025-04-22
feat: introduce ucs06 funded dispatch

hussein-aitlahcen:


cor:
@hussein-aitlahcen is this ready for review or still a draft? There is a docs page created in this PR, but it is still empty so difficult to give the protocol a good review without description of what it should do

cosmwasm/ibc-union/app/ucs06-funded-dispatch/src/types.rs
```diff
@@ -0,0 +1,50 @@
...
+    struct FundedDispatchFund {
+        bytes token;
+        uint256 amount;
+    }
+
+    #[derive(Debug)]
+    struct FundedDispatchParameters {
+        uint256 flags;
+        FundedDispatchFund[] funds;
+        bytes contractAddress;
```
aeryz, line 18: NIT: Make these fields snake case to be consistent with ZKGM and Rust conventions

cosmwasm/ibc-union/app/ucs06-funded-dispatch/src/contract.rs
```diff
@@ -0,0 +1,219 @@
...
+    for fund in params.funds.iter() {
+        let denom = std::str::from_utf8(&fund.token).map_err(|_| ContractError::InvalidDenom {
+            denom: fund.token.to_vec().into(),
+        })?;
+        let amount = u128::try_from(fund.amount)
+            .map_err(|_| ContractError::InvalidAmount {
+                amount: fund.amount,
+            })?
+            .into();
+        if deps.querier.query_wasm_contract_info(denom).is_ok() {
```
aeryz, line 96: would be better if we query cw20 token info to make sure that the denom is cw20. It's still fine to do it this way though since `IncreaseAllowance` will fail anyways otherwise.

aeryz (approved):

cosmwasm/ibc-union/app/ucs06-funded-dispatch/src/contract.rs
hussein-aitlahcen: I think it's best to use the increase allowance because it's then the only interface required for a contract to be able to be used. It means the contract could be custom (not cw20) and still be used in the process.

aeryz (approved):

PR #4420, opened 2025-04-24
feat(deployments): add a chain to test

Swepool:


deployments/editions.json
```diff
@@ -17,6 +17,9 @@
   },
   "btc": {
     "chains": {
+      "union.union-1": {
```
qlp, line 20: should be sorted: run `jq . editions.json -S | sponge editions.json`

(there's a pending pr explaining this)

btw: before moving to pro, you'll probably want all btc values to be `production`

PR #4432, opened 2025-04-25
no bucket checks for testnets

Swepool:


app2/src/lib/transfer/shared/components/ChainAsset/ChainSelector.svelte
```diff
@@ -171,6 +171,9 @@ const filterChainsByTokenAvailability = (chains: Array<Chain>): Array<ChainWithA
             onNone: () => chains.map(chain => Tuple.make(chain, false)),
             onSome: ({ baseToken, sourceChain }) =>
               chains.map(destinationChain => {
+                if (destinationChain.testnet === true) {
+                  return Tuple.make(destinationChain, true)
+                }
```
cor, line 176: Can you add a small comment to explain that for testnets we don't aply bucket-based ratelimiting?

cor (changes requested):

cor (approved):

PR #4499, opened 2025-05-01
chore(docs): display chain name and chain ID

PoisonPhang:


docs/src/components/svelte/tables/ContractDeploymentsTable.svelte
```diff
@@ -5,7 +5,7 @@ import deployments from "~root/deployments/deployments.json" with { type: "json"
 
 {#each deployments as network}
   <h3>
-    {network.chain_id} Deployments
+    {network.universal_chain_id} Deployments
```
benluelo, line 8: we can ditch the "Deployments" here, since the page is called deployments

benluelo (approved):

PR #4516, opened 2025-05-05
chore(ts-sdk): added cosmos-evm read with block/height

Caglankaan:


ts-sdk/src/cosmos/extended-cw-client.ts
```diff
@@ -0,0 +1,61 @@
...
+      throw new Error(`HTTP ${resp.status}: ${JSON.stringify(resp.data)}`)
+    }
+
+    return resp.data
+  }
+
+  async getBalanceAtHeight(address: string, denom: string, height: number) {
+    const base = this.restUrl
+    const url = `${base}/cosmos/bank/v1beta1/balances/${address}`
+    const resp = await fetch(url, {
```
cor, line 41: i recommend not going for this approach of extending `CosmWasmClient`, at all and to instead just write simple functions that make the Http call and return an Effect with proper error types rather than throwing errors. Using the client here is basically  useless because all that we're doing is just extracting the restURL from the client and then making a simple http request. Because we're doing it this way we lose valuable error information as we're not using our Effectful methods of making http calls that we have elsewhere in the sdk

cor (changes requested):
I think we should not extend CosmWasmClient, and instead just write simple functions that return an Effect. All we're doing is making an HTTP call

PR #4518, opened 2025-05-05
feat(zkgm): allow native eth bridging through eip 7528 address

hussein-aitlahcen:
- make `send` function payable and wrap any sent eth if an order for base token https://eips.ethereum.org/EIPS/eip-7528 is present
- handle unwrapping of weth back to eth on protocol filling in unescrow branch


evm/contracts/apps/ucs/03-zkgm/Lib.sol
```diff
@@ -36,7 +36,7 @@ library ZkgmLib {
         0xC0DE00000000000000000000000000000000000000000000000000000000BABE;
 
     address public constant NATIVE_ETH_MAGIC =
```
benluelo, line 38: maybe rename to `ERC_7528_ETH_ADDRESS` or something

Sammie-Wealth:
Absoluteluy yall,Good morning yall.

benluelo (approved):

PR #4530, opened 2025-05-08
fix(cosmos-client): fee market rounding

PoisonPhang:
- fixes fee market rounding
- fixes deployments on devnet-eth


cor (approved):

Maciejrybak4 (commented):
Maciej rybak4

PR #4536, opened 2025-05-09
feat(hubble): support instantiate height specification in token list

qlp:


hubble/src/token_fetcher/client.rs
```diff
@@ -86,6 +87,49 @@ impl Display for Token {
...
+            f.write_fmt(format_args!("hubble: {}", hubble))
+        } else {
+            f.write_str("no extensions")
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Deserialize)]
+pub struct HubbleExtension {
+    #[serde(default, deserialize_with = "deserialize_opt_u64_from_string")]
```
benluelo, line 107: we have this in serde_utils already if you want to use that crate

benluelo (approved):

hubble/src/token_fetcher/client.rs
qlp: I only see u64_hex? or is there some generic deserializer I can use?

benluelo: https://github.com/unionlabs/union/blob/3fc7119b562776d14f1718705197758259d7a31e/lib/serde-utils/src/lib.rs#L626

qlp: fixed

benluelo (approved):

PR #4538, opened 2025-05-10
feat(app2): introduce Effect runtime

ehegnes:
Previously we had no way to provision Effect requirements in an app-wide manner. This changeset introduces a [`ManagedRuntime`](https://effect.website/docs/runtime/#managedruntime) as a module singleton.

The runtime is guaranteed to exist upon page load by virtue of SvelteKit's [`init`](https://svelte.dev/docs/kit/hooks#Shared-hooks-init) hook.

Initial integration targets the following features:
- Integration of a [custom Effect `Logger`](https://effect.website/docs/observability/logging/#custom-loggers) for Datadog interoperability.
- Provisioning of runtime to the in-house Effect "runes" which handle automatic resource cleanup and interruption (see #4444).
- Minimal dependency inversion exchanging a "Live" layer with a "Test" layer for integration testing when running via `vitest`.


eyestorefivemstore:
The performance optimizations here are quite clever. Nice engineering work!

flake.nix
```diff
@@ -240,12 +240,18 @@
           pkgsUnstable = import inputs.nixpkgs-unstable { inherit system; };
 
           gitRev = if (builtins.hasAttr "rev" self) then self.rev else "dirty";
+          gitShortRev = self.shortRev or (self.dirtyShortRev or "dirty");
+          lastModified = if self ? lastModified then builtins.toString self.lastModified else "0";
+          lastModifiedDate = self.lastModifiedDate or "1970-01-01T00:00:00Z";
```
ehegnes, line 245: Additional flake metadata for embedding in `app2` builds.

@benluelo I can not cast `lastModified` (normally an integer representing seconds) if you would like to use this elsewhere.

ehegnes (commented):
Annotated anything notable.

All else is replacing usages of `Effect.run*` with the exports from `"$lib/runtime"`.

flake.nix
benluelo: wdym you can't cast it?

ehegnes: I can, just casted here because lazy. Eh, I'll adjust to integer anyway so it's an expected result and cast to string only in the `app2` derivation.

cor (approved):

PR #4541, opened 2025-05-10
 Add missing schedule trigger to Nightly workflow

Budalebah:
This PR adds a schedule trigger to the Nightly workflow. The workflow contains a job (create-issue) that runs only when the event is schedule and the build fails. However, without defining a schedule trigger, this condition can never be met.


PoisonPhang:
Thank you, but we are intentionally not running this job atm as we don't utilize nixbuild.net anymore.

PR #4551, opened 2025-05-13
feat(uniond): cosmwasm metrics

PoisonPhang:


Maciejrybak4 (commented):
maciek.rybak21@op.pl

PoisonPhang:
Determined that CosmWasm metrics are not functioning as expected and will not attempt to add them this way.

PR #4564, opened 2025-05-18
feat(chain-kitchen): init

benluelo:
wip


cor (approved):

lib/chain-kitchen/src/lib.rs
```diff
@@ -0,0 +1,241 @@
...
+}
+
+#[derive(Debug, PartialEq, Eq)]
+#[repr(transparent)]
+pub struct Protocol(str);
+
+impl Protocol {
+    pub const RPC: &'static Self = Self::new("rpc").unwrap();
+    pub const REST: &'static Self = Self::new("rest").unwrap();
+    pub const GRPC: &'static Self = Self::new("grpc").unwrap();
```
qlp, line 85: we don't do grpc anymore. we could replace it with `evm-rpc`

PR #4582, opened 2025-05-21
Add mainnet-only switch to transfers and packets

cor:
- **chore(app2): update graphql schema**
- **feat(app2): add option to only query mainnet transfers**
- **fix(app2): use shortcuts that also exist on linux**
- **fix(app2): better edition defaults system**
- **feat(app2): introduce switch component**
- **fix(app2): switch events and reactivity**
- **refactor(app2): use svelte5 event callbacks**
- **feat(app2): mainnet only for packet explorer**

### DIFF
--- a/app2/src/generated/graphql-env.d.ts
+++ b/app2/src/generated/graphql-env.d.ts
@@ -1,232 +1,6 @@
 /* eslint-disable */
 /* prettier-ignore */
 
-export type introspection_types = {
-    'Boolean': unknown;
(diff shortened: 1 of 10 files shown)


Swepool (dismissed):
Looks good, u sure we want to use ctrl for shortcut tho?

ehegnes (approved):

ehegnes (approved):

ehegnes:
> Looks good, u sure we want to use ctrl for shortcut tho?

IMHO better than mod key which usually hooks into windowing system shortcuts and could collide.

PR #4584, opened 2025-05-21
feat(ibc-union): add cw-orch

benluelo:
revival of #3508, which contained a merge commit, and rather than dealing with rebasing that i just cherrypicked the changes.

### DIFF
--- a/cosmwasm/ibc-union/core/msg/src/query.rs
+++ b/cosmwasm/ibc-union/core/msg/src/query.rs
@@ -1,20 +1,37 @@
 use ibc_union_spec::{ChannelId, ClientId, ConnectionId};
 use unionlabs_primitives::H256;
-
-#[derive(serde::Serialize, serde::Deserialize)]
+#[derive(serde::Serialize, serde::Deserialize, Debug)]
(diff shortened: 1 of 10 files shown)


cosmwasm/ibc-union/core/msg/src/query.rs
```diff
@@ -1,20 +1,37 @@
...
-#[derive(serde::Serialize, serde::Deserialize)]
+#[derive(serde::Serialize, serde::Deserialize, Debug)]
 #[serde(deny_unknown_fields, rename_all = "snake_case")]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+#[cfg_attr(
+    feature = "cw-orch-interface",
+    derive(cosmwasm_schema::QueryResponses, cw_orch::QueryFns)
+)]
 pub enum QueryMsg {
+    #[cfg_attr(feature = "cw-orch-interface", returns(ibc_union_spec::Timestamp))]
```
benluelo, line 11: @adairrr is there any way to guarantee this return type is valid?

PoisonPhang (approved):

PR #4593, opened 2025-05-22
chore(ci): disable e2e tests and external pkgs

PoisonPhang:


cor (approved):

garnix.yaml
```diff
@@ -69,6 +71,8 @@ builds:
       - 'homeConfigurations.*'
       - 'nixosConfigurations.*'
     exclude:
+      - 'checks.*.all-crates-buildable-individually'
```
benluelo, line 74: we should run this in nightly

benluelo (approved):

garnix.yaml
cor: agree

PR #4609, opened 2025-05-24
feat(app2): add client-side GQL caching with TTL 

ehegnes:


app2/src/lib/graphql/error.ts
```diff
@@ -0,0 +1,20 @@
+import { Schema as S } from "effect"
+import type { ClientError } from "graphql-request"
+
+export class GraphQLError extends S.TaggedError<GraphQLError>("GraphQLError")("GraphQLError", {
```
ehegnes, line 4: So we can get meaningful errors from GraphQL rather than `UnknownException`.

app2/src/lib/graphql/service.ts
```diff
@@ -0,0 +1,215 @@
...
+import { GraphQLError } from "./error"
+
+/**
+ * Define TTL strategy given some GQL document.
+ * NOTE: Determines TTL by virtue of custom `@cached` directive.
+ * TODO: Should probably define full cache behavior.
+ */
+class GraphQLCache extends Effect.Service<GraphQLCache>()("app/GraphQL/Cache", {
+  effect: Effect.gen(function*() {
+    const timeToLive = <
```
ehegnes, line 32: Determine TTL by traversing `DocumentNode` AST for `@cached` directive.

app2/src/lib/services/cosmos/balances.ts
```diff
@@ -122,5 +122,4 @@ export const fetchCosmosBalance = ({
     }),
     Effect.scoped,
     Effect.provide(FetchHttpClient.layer),
-    withTracerDisabledWhen(() => true), // important! this prevents CORS issues: https://github.com/Effect-TS/effect/issues/4568
```
ehegnes, line 125: Moved so it can be provided directly to the `HttpClient` in `fetchDecode`.

app2/src/lib/utils/queries.ts
```diff
@@ -29,36 +33,25 @@ export const fetchDecode = <S>(schema: Schema.Schema<S>, url: string) =>
...
-    document,
-    operationNamesFromDocumentNode,
-    A.head,
-    Option.getOrElse(() => "unknown"),
+): Effect.Effect<S, FetchDecodeGraphqlError, GraphQL> =>
+  Effect.andThen(
+    GraphQL,
+    ({ fetch }) =>
+      pipe(
+        fetch(new GraphQLRequest({ document, variables })),
```
ehegnes, line 50: `GraphQL` service is only be responsible for querying, not schema transforms.

This may change if/when we migrate to declaring requests/results with more `Schema` classes.

app2/src/lib/runtime.ts
```diff
@@ -1,10 +1,13 @@
...
+import type { GraphqlQuery } from "@aptos-labs/ts-sdk"
+import { Effect, Layer, ManagedRuntime, Match, pipe } from "effect"
 import { isNotUndefined } from "effect/Predicate"
+import type { GraphQL } from "./graphql/service"
 
 const IS_VITEST = isNotUndefined(import.meta.vitest)
 
-type AppLayer = Layer.Layer<never, never, never>
+type AppLayer = Layer.Layer<GraphQL, never, never>
+export type AppContext = Layer.Layer.Success<AppLayer>
```
ehegnes, line 10: Our runtime now provisions the `GraphQL` service for all effects.

ehegnes:
This service should arguably move into the SDK, but a strategy for dependency inversion of the GraphQL URL is required, and migrating all app GQL usage to SDK is out-of-scope.

cor (approved):

Issue #4360, opened 2025-04-16
write a tool to sort rust files alphabetically


benluelo:
see https://github.com/rust-lang/rustfmt/issues/2359#issuecomment-1942559776 for why. should be trivial to do with syn, just ensure doc comments are retained (shouldn't require any special handling since they parse as `#[doc]` attributes anyways).

benluelo:
updated the description @marvelshan, let me know if you have any questions

benluelo:
you can add `experimental-features = nix-command flakes` to your `nix.conf` (mine is located at `/etc/nix/nix.conf`). how you set this (either manually or via a `nixos-rebuild`) depends on how you've installed nix

marvelshan:
@benluelo I’ve found two issues that I’m interested in working on: #3393 and #3396. Could you please advise which issue would be better for me to start with? Or, do you have any other recommendations for issues that would be suitable for me to work on?

PR #4614, opened 2025-05-26
feat(tools): rustfmt sort tool

marvelshan:
## Which issue does this PR close?
closes https://github.com/unionlabs/union/issues/4360

## Why
This change introduces a deterministic Rust file sorting tool (rustfmt-sort) into the proto code generation pipeline. The motivation is to ensure that all generated Rust files have a stable, alphabetically sorted item order, which:

- Reduces unnecessary diffs in generated code,
- Improves reproducibility and code review experience,
- Makes it easier to track meaningful changes in generated files. This is especially important for large proto-generated codebases where minor changes in proto definitions or generation order can otherwise cause noisy diffs.
  Additionally, due to limitations in the proto generator, a manual patch for the timestamp field in CanonicalVote is still required after generation.

## Implementation details
- Uses syn to parse and mutate the Rust syntax tree, and quote to regenerate the code.
- Recursively sorts items in modules as well.
- The tool is invoked automatically as part of the proto generation process.

## Modules Affected
- tools/rustfmt-sort
- tools/rust-proto.nix
- tools/tools.nix
- Cargo.lock
- Cargo.toml

### DIFF
--- a/generated/rust/protos/src/amino.rs
+++ b/generated/rust/protos/src/amino.rs
@@ -1,2 +1,1 @@
-// @generated
-// @@protoc_insertion_point(module)
+
--- a/generated/rust/protos/src/babylon.btccheckpoint.v1.rs
+++ b/generated/rust/protos/src/babylon.btccheckpoint.v1.rs
@@ -1,4 +1,104 @@
-// @generated
+/// BTCCheckpointInfo contains all data about best submission of checkpoint for
+/// given epoch. Best submission is the submission which is deeper in btc ledger
+#[allow(clippy::derive_partial_eq_without_eq)]
+#[derive(Clone, PartialEq, :: prost :: Message)]
+pub struct BtcCheckpointInfo {
+    /// epoch number of this checkpoint
+    #[prost(uint64, tag = "1")]
+    pub epoch_number: u64,
+    /// btc height of the best submission of the epoch
+    #[prost(uint32, tag = "2")]
+    pub best_submission_btc_block_height: u32,
+    /// hash of the btc block which determines checkpoint btc block height i.e.
+    /// youngest block of best submission
+    #[prost(bytes = "vec", tag = "3")]
+    pub best_submission_btc_block_hash: ::prost::alloc::vec::Vec<u8>,
+    /// the BTC checkpoint transactions of the best submission
+    #[prost(message, repeated, tag = "4")]
+    pub best_submission_transactions: ::prost::alloc::vec::Vec<TransactionInfo>,
+    /// list of vigilantes' addresses of the best submission
+    #[prost(message, repeated, tag = "5")]
+    pub best_submission_vigilante_address_list: ::prost::alloc::vec::Vec<CheckpointAddresses>,
+}
+/// BTCCheckpointInfoResponse contains all data about best submission of checkpoint for
+/// given epoch. Best submission is the submission which is deeper in btc ledger.
+#[allow(clippy::derive_partial_eq_without_eq)]
+#[derive(Clone, PartialEq, :: prost :: Message)]
+pub struct BtcCheckpointInfoResponse {
+    /// EpochNumber of this checkpoint.
+    #[prost(uint64, tag = "1")]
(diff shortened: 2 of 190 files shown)


tools/rustfmt-sort/src/main.rs
```diff
@@ -0,0 +1,44 @@
+use std::{env, fs};
+
+use quote::ToTokens;
+use syn::{visit_mut::VisitMut, Item};
+
+fn main() {
+    let mut file =
+        syn::parse_file(&fs::read_to_string(env::args().nth(1).unwrap()).unwrap()).unwrap();
```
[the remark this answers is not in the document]
benluelo: you can use clap like we use in the rest of our binaries

tools/rustfmt-sort/src/main.rs
```diff
@@ -0,0 +1,56 @@
+use std::fs;
+
+use clap::Parser;
+use quote::ToTokens;
+use syn::{visit_mut::VisitMut, Item};
+
+#[derive(Parser, Debug)]
+#[command(name = "rustfmt-sort", about = "Sort Rust items in a file")]
+struct Args {
+    #[arg(value_name = "FILE")]
```
benluelo, line 10: no need for `value_name`, the entire attribute can be removed

marvelshan:
@benluelo I tried running the CI for it but ran into a couple of issues that I’m currently stuck on, and I was wondering if you might have any suggestions.

Here are the main errors I encountered:

1. `dprint` formatting failed:
    
    ```
    Error resolving plugin /nix/store/.../dprint-plugin-typescript-0.95.4: Is a directory (os error 21)
    ```
    
    It seems like `dprint` is failing to resolve the TypeScript plugin from the Nix store — possibly due to a misreference to a directory instead of a plugin file?
    
2. `rustfmt` errors related to `gen` keyword:
    
    ```
    error: expected identifier, found reserved keyword `gen`
    ```
    
    This occurs in multiple files using `schemars::gen::SchemaGenerator`, where `gen` appears to be conflicting with the new `gen` keyword in Rust 2024 edition. I assume the fix is to rename it to `r#gen`, but I’m not entirely sure if that’s the expected approach here.
    
Would you happen to know what might be causing the `dprint` plugin resolution issue, and do you have any suggestions on how best to address the `gen` keyword conflicts for `rustfmt`?
Any pointers would be really appreciated. Thank you!

flake.lock
benluelo, line 1: revert all changes to this file

marvelshan: OK, I got it.

marvelshan:
@benluelo I'm using the Rust version provided by the repo via `nix develop`, and I'm inside the dev shell:

```bash
(nix:union-devShell-env) zaki@ubuntu:~/union$ echo $IN_NIX_SHELL
impure

(nix:union-devShell-env) zaki@ubuntu:~/union$ rustc --version
rustc 1.87.0-nightly (3f5502370 2025-03-27)

(nix:union-devShell-env) zaki@ubuntu:~/union$ cargo --version
cargo 1.87.0-nightly (a6c604d1b 2025-03-26)
```

The edition field in `Cargo.toml` is still set to `"2021"`:

```toml
[workspace.package]
edition = "2021"
```

However, when I run `nix fmt`, I still get formatting errors that seem to assume edition 2024 is in use:

```
ERRO formatter | rustfmt: failed to apply with options '[--config skip_children=true --edition 2024]': exit status 1

error: expected identifier, found reserved keyword `gen`
```

This happens in multiple files, mostly in code using `schemars::gen::SchemaGenerator`, where `gen` is being flagged as a reserved keyword (which is new in edition 2024).

Also, I noticed `nix fmt` prints the following:

```
evaluation warning: crane requires at least nixpkgs-25.05, supplied nixpkgs-24.11
```

Could this be causing `rustfmt` to use the wrong edition somehow? Let me know if there’s something I should adjust in my setup. Thanks!

benluelo:
you've updated the flake.lock again. revert changes to that file, and don't update it.

PR #4625, opened 2025-05-28
feat(app): gas

ehegnes:


app2/src/lib/transfer/normal/steps/FillingStep.svelte
```diff
@@ -32,6 +31,38 @@ const {
...
+
+const realPriceDisplay = $derived(pipe(
+  realPrice.current,
+  Option.map(Exit.match({
+    onFailure: Cause.prettyErrors,
+    onSuccess: (x) => JSON.stringify(x, null, 2),
+  })),
+  Option.getOrElse(() => "Loading price..."),
+))
+
```
ehegnes, line 65: Example usage in a component, leveraging our custom Effect runes.

app2/src/lib/transfer/normal/steps/FillingStep.svelte
```diff
@@ -32,6 +31,38 @@ const {
...
+  fakePrice.current,
+  Option.map(Exit.match({
+    onFailure: Cause.prettyErrors,
+    onSuccess: (x) => JSON.stringify(x, null, 2),
+  })),
+  Option.getOrElse(() => "Loading price..."),
+))
+
+const realPrice = runPromiseExit$(
+  Effect.provide(eff, PriceOracle.PriceOracle.Pyth),
```
ehegnes, line 54: Providing a layer or service breaks reactivity such that the effect only runs once (until HMR triggers).

Would be amazing to solve this, but it shouldn't impact us in the short-term given we will only use the global app runtime in practice.

PR #4679, opened 2025-06-11
feat(hubble): nats - decouple block processing

qlp:
decouple reading chain data from writing into database by introducing nats event streams

### DIFF
--- a/hubble/src/indexer/consumer.rs
+++ b/hubble/src/indexer/consumer.rs
@@ -28,6 +28,7 @@ use crate::{
         },
     },
     postgres::{fetch_internal_chain_id_for_universal_chain_id, schedule_replication_reset},
+    utils::human_readable::human_readable_bytes,
 };
 
 pub struct BlockUpdate {
@@ -460,22 +461,5 @@ fn get_message_meta(message: &async_nats::jetstream::Message) -> Result<MessageM
             nats_stream_sequence,
             nats_consumer_sequence,
         ))
-    }
-}
-
-fn human_readable_bytes(bytes: usize) -> String {
-    const UNITS: [&str; 4] = ["B", "KiB", "MiB", "GiB"];
-    let mut size = bytes as f64;
-    let mut unit = 0;
-
-    while size >= 1024.0 && unit < UNITS.len() - 1 {
-        size /= 1024.0;
-        unit += 1;
-    }
-
-    if unit == 0 {
-        format!("{} {}", bytes, UNITS[unit])
-    } else {
-        format!("{:.1} {}", size, UNITS[unit])
     }
 }
--- a/hubble/src/main.rs
+++ b/hubble/src/main.rs
@@ -20,6 +20,7 @@ mod metrics;
 mod postgres;
 mod race_client;
 mod token_fetcher;
+mod utils;
 
 #[cfg(not(target_env = "msvc"))]
 use tikv_jemallocator::Jemalloc;
--- a/hubble/src/utils/human_readable.rs
+++ b/hubble/src/utils/human_readable.rs
@@ -1,0 +1,16 @@
+pub fn human_readable_bytes(bytes: usize) -> String {
+    const UNITS: [&str; 4] = ["B", "KiB", "MiB", "GiB"];
+    let mut size = bytes as f64;
+    let mut unit = 0;
+
+    while size >= 1024.0 && unit < UNITS.len() - 1 {
+        size /= 1024.0;
+        unit += 1;
+    }
+
+    if unit == 0 {
+        format!("{} {}", bytes, UNITS[unit])
+    } else {
+        format!("{:.1} {}", size, UNITS[unit])
+    }
+}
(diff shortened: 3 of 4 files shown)


hubble/src/indexer/consumer.rs
```diff
@@ -0,0 +1,482 @@
...
+    .await?;
+
+    Ok(())
+}
+
+// what should we do with a block at a specific height.
+enum Action<'a> {
+    Delete(&'a MessageMeta, &'a BlockUpdate),
+    Update(&'a MessageMeta, &'a BlockUpdate, &'a Vec<&'a BlockEvent>), // events are guaranteed to belong to the same block height
+    Insert(&'a MessageMeta, BlockUpdate, &'a Vec<&'a BlockEvent>),
```
cor, line 317: missed lifetime &’a for blockupdate here?

hubble/src/indexer/consumer.rs
```diff
@@ -0,0 +1,482 @@
...
+        size /= 1024.0;
+        unit += 1;
+    }
+
+    if unit == 0 {
+        format!("{} {}", bytes, UNITS[unit])
+    } else {
+        format!("{:.1} {}", size, UNITS[unit])
+    }
+}
```
cor: Is there no (standard) lib function for that? seems like something that should’ve been implemented a ton of times before

hubble/src/indexer/event.rs
```diff
@@ -0,0 +1,324 @@
...
+use crate::indexer::{
+    api::{BlockHeight, BlockRange, BlockReference, FetcherClient, IndexerError, UniversalChainId},
+    nats::subject_for_block,
+    postgres::nats::schedule,
+    Indexer,
+};
+
+#[derive(serde::Serialize, serde::Deserialize)]
+pub struct HubbleEvent {
+    pub version: u8,
```
cor, line 27: how is backwards/forwards compatibility defined with this versioning system? does perhaps some sort of semver system make sense for if a producer starts providing extra data that is _optionally_ consumed and not a breaking update?

hubble/src/indexer/event.rs
```diff
@@ -0,0 +1,324 @@
...
+
+impl From<&BlockReference> for Range {
+    fn from(value: &BlockReference) -> Self {
+        Self {
+            start_inclusive: value.height,
+            end_exclusive: value.height + 1,
+        }
+    }
+}
+
```
cor: why does the first one not have `+ 1` but the second one does have `+ 1`? _If_ this is intentional, i recommend providing doc comments for the difference

hubble/src/indexer/mod.rs
```diff
@@ -98,6 +220,69 @@ impl Default for FinalizerConfig {
...
+        Duration::from_secs(5)
+    }
+
+    pub fn default_retry_error_sleep() -> Duration {
+        Duration::from_secs(5)
+    }
+
+    pub fn default_max_blocks_in_message() -> u64 {
+        1000
+    }
```
cor, line 269: I would just inline these in the `impl Default for FixerConfig` as it’s realistically the only place where they’ll be used. less bloat

hubble/src/indexer/publisher.rs
```diff
@@ -0,0 +1,78 @@
...
+
+enum PublisherLoopResult {
+    RunAgain,
+    TryAgainLater,
+}
+
+impl<T: FetcherClient> Indexer<T> {
+    pub async fn run_publisher(&self) -> Result<(), IndexerError> {
+        if let Some(nats) = &self.nats {
+            loop {
```
cor: I would use `let else` here to flatten the structure

cor (changes requested):
Looks great! nats integration very cool. a few minor comments

hubble/src/indexer/consumer.rs
qlp: No, this is easier, because for delete/update the origin is a vec we read from db, and for insert it's a new one based on the message we read.

hubble/src/indexer/consumer.rs
qlp: ah, yes. but reminded the discussion we had on adding a dependency for 5 lines of code. can add a lib too. lmk

hubble/src/indexer/event.rs
qlp: not specified yet, it's mostly a tag that we can use to handle change. imo we should not process messages if we're not supporting the version (so consumers should nack them, but should be backward compatible).

hubble/src/indexer/event.rs
qlp: a range for a single block starts at the block height (inclusive) and ends one block after the block height (because it's exclusive).

i'll add doc

hubble/src/indexer/mod.rs
qlp: it's used twice: in serde config _and_ Default implementation

hubble/src/indexer/consumer.rs
cor: makes sense, ty

cor (approved):

PR #4681, opened 2025-06-11
feat(parlia): implement misbehaviour and expiry

benluelo:


cosmwasm/ibc-union/lightclient/parlia/src/client.rs
```diff
@@ -168,22 +182,64 @@ impl IbcClient for ParliaLightClient {
...
+        ctx: IbcClientCtx<Self>,
         _caller: Addr,
-        _misbehaviour: Self::Misbehaviour,
+        misbehaviour: Self::Misbehaviour,
         _relayer: Addr,
     ) -> Result<Self::ClientState, IbcClientError<Self>> {
-        Err(Error::Unimplemented.into())
+        let ClientState::V1(client_state) = ctx.read_self_client_state()?;
+
+        if misbehaviour.attestation_1.number != misbehaviour.attestation_2.number {
```
aeryz, line 192: we must check whether the attestation or target whatever is not the same thing. otherwise one can provide a valid misbehaviour by giving the same data twice.

PR #4737, opened 2025-06-26
feat(uniond): add union staking (proof of possession)

PoisonPhang:


hussein-aitlahcen:
are we planning to build the migration in a second PR?

PoisonPhang:
> are we planning to build the migration in a second PR?

Yes, these changes are *mostly* trivial. The migration should be reviewed separately.

This will also enable running the devnet without PoA, which may be useful to some testing.

PR #4778, opened 2025-07-05
fix: Prevent Recursive Call in Update module.go

moshderte:
Summary
This PR fixes a recursive call bug in the RegisterInterfaces method of the AppModuleBasic struct.

What was the issue?
Previously, the method was defined as: func (AppModuleBasic) RegisterInterfaces(registry codectypes.InterfaceRegistry) {
    RegisterInterfaces(registry)
This leads to an infinite recursive call, as the method was calling itself instead of the intended package-level RegisterInterfaces function. This results in a stack overflow panic at runtime.

What was changed?
Renamed the package-level RegisterInterfaces function to RegisterCometBLSInterfaces and updated the method accordingly: func (AppModuleBasic) RegisterInterfaces(registry codectypes.InterfaceRegistry) {
    RegisterCometBLSInterfaces(registry)

Why is this important?
Prevents runtime panic due to infinite recursion

Makes the code more maintainable and less prone to name shadowing errors


onyedika12:
Fixed

moshderte:
@aeryz check please

PR #4781, opened 2025-07-08
feat(uniond): v1.2.0 migration

PoisonPhang:
- migration for v1.2
- Mints token
- Changes default token in all contexts


demiGOD-web:
💞💞💞💞💞💞💞💞💞💞💞💞💞
Keep building sers💪🏾

hussein-aitlahcen (dismissed):
Good, let's have multiple eyes review this and prepare a testnet upgrade. BTW, for UNION_FOUNDATION_MULTI_SIG lets switch on the chain ID for the upgrade such that we send the supply on the whale account in testnet, and the foundation multisig in mainent

uniond/app/upgrades/v1_2_0/upgrade.go
```diff
@@ -0,0 +1,240 @@
...
+	minttypes "github.com/cosmos/cosmos-sdk/x/mint/types"
+	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
+
+	"github.com/unionlabs/union/uniond/app/upgrades"
+)
+
+const U_BASE_DENOM = "au"
+
+// One U in `au`
+const ONE_U = 1000000000000000000
```
benluelo: maybe better to write this as 10 * 10^18 if possible?

uniond/app/upgrades/v1_2_0/upgrade.go
```diff
@@ -0,0 +1,240 @@
...
+
+// Total supply of U (note, not in `au`)
+const U_TOTAL_SUPPLY = 10000000000000
```
benluelo: same comment here

uniond/app/upgrades/v1_2_0/upgrade.go
```diff
@@ -0,0 +1,240 @@
...
+
+// Union foundation multisig address
+const UNION_FOUNDATION_MULTI_SIG = "union1cpz5fhesgjcv2q0640uxtyur5ju65av6r8fem0"
```
benluelo: note that we need to add a branch for testnet vs mainnet here before we upgrade mainnet

uniond/app/upgrades/v1_2_0/upgrade.go
```diff
@@ -0,0 +1,240 @@
...
+		govParams.ExpeditedMinDeposit = []sdk.Coin{getUFromU64(50)}
+		err = keepers.GovKeeper.Params.Set(ctx, govParams)
+		if err != nil {
+			return nil, err
+		}
+
+		// Update x/crisis
+		_, err = keepers.CrisisKeeper.UpdateParams(ctx, &crisistypes.MsgUpdateParams{
+			Authority:   keepers.GovKeeper.GetAuthority(),
+			ConstantFee: getBaseUFromString("1000"),
```
benluelo: where did this value come from/ what does it mean?

benluelo (commented):
just a few questions and nits

uniond/app/upgrades/v1_2_0/upgrade.go
PoisonPhang: This is what it's set to on both networks genesis.

> Due to the anticipated large gas cost requirement to verify an invariant (and potential to exceed the maximum allowable block gas limit) a constant fee is used instead of the standard gas consumption method. The constant fee is intended to be larger than the anticipated gas cost of running the invariant with the standard gas consumption method.

We can update it to be higher if need be.

uniond/app/upgrades/v1_2_0/upgrade.go
PoisonPhang: Are you saying you want this value in au?
Also, this should be 10 billion, my bad.

e2e/upgrades.nix
```diff
@@ -67,6 +67,9 @@ in
       union.wait_until_succeeds('[[ $(curl "http://localhost:26657/block" --fail --silent | ${pkgs.lib.meta.getExe pkgs.jq} ".result.block.header.height | tonumber > 1") == "true" ]]')
 
       ${upgradeTo "v1.1.0" 10}
+      ${upgradeTo "v1.2.0" 20}
+      union.wait_until_succeeds('[[ $(curl "http://localhost:26660/block" --fail --silent | ${pkgs.lib.meta.getExe pkgs.jq} ".result.block.header.height | tonumber > 40") == "true" ]]', timeout=120)
```
benluelo, line 71: it would be good if we could also verify the contents of the bank module and the delegations here

uniond/app/upgrades/v1_2_0/upgrade.go
```diff
@@ -0,0 +1,240 @@
...
+const U_BASE_DENOM = "au"
+
+// One U in `au`
+const ONE_U = 1_000_000_000_000_000_000
+
+// Total supply of U (note, not in `au`)
+const U_TOTAL_SUPPLY = 10_000_000_000
+
+// Union foundation multisig address
+const UNION_FOUNDATION_MULTI_SIG = "union1cpz5fhesgjcv2q0640uxtyur5ju65av6r8fem0" // NOTE: Must add branch logic for mainnet before running migration
```
benluelo: on second thought, maybe it'd be best to keep these in a map indexed by chain id?

uniond/app/upgrades/v1_2_0/upgrade.go
benluelo: also why not just use getUFromU64 here lol

hussein-aitlahcen (approved):

cor (approved):

PR #4859, opened 2025-07-19
feat(lightclient): add CommitClientStatus message to persist client s…

### LINKED ISSUE #4842: Client status commitment
The light clients can be in 3 state at all times:

1. Active
2. Expired
3. Frozen

The light client is only usable at the `Active` state. And this status matters the most for the `update_client` function because we definitely don't want to add a new state to a frozen or expired client.

Currently, if the client is frozen, we write it under `ClientState` of the corresponding client. And if its expired, we don't commit it at all since it is calculated dynamically at the time where the `GetStatus` query is called. We want to create a commitment for the clients where we commit the status as well. This is going to be very useful for the state lens clients since they will need to guarantee that the client is not frozen.

Right after a call to the `misbehaviour`, `update_client`, and right after the execution of `migrate_state`, we should query the status of the client and commit it since these are the only entrypoints where we might have a state change.

0xumutkk:
Adds a new 'CommitClientStatus' message to explicitly commit the client's status (Active, Expired, or Frozen) into state storage as a hashed commitment using 'store_commit'.

Closes #4842

### DIFF
--- a/cosmwasm/ibc-union/core/src/contract.rs
+++ b/cosmwasm/ibc-union/core/src/contract.rs
@@ -28,7 +28,8 @@ use ibc_union_msg::{
 use ibc_union_spec::{
     path::{
         commit_packets, BatchPacketsPath, BatchReceiptsPath, ChannelPath, ClientStatePath,
-        ConnectionPath, ConsensusStatePath, COMMITMENT_MAGIC, COMMITMENT_MAGIC_ACK,
+        ClientStatusPath, ConnectionPath, ConsensusStatePath, COMMITMENT_MAGIC,
+        COMMITMENT_MAGIC_ACK,
     },
     Channel, ChannelId, ChannelState, ClientId, Connection, ConnectionId, ConnectionState,
     MustBeZero, Packet, Status, Timestamp,
@@ -42,9 +43,8 @@ use unionlabs::{
 use crate::{
     state::{
         ChannelOwner, Channels, ClientConsensusStates, ClientImpls, ClientRegistry, ClientStates,
-        ClientStatuses, ClientStore, ClientTypes, Commitments, Connections, ContractChannels,
-        NextChannelId, NextClientId, NextConnectionId, QueryStore, WhitelistedRelayers,
-        WhitelistedRelayersAdmin,
+        ClientStore, ClientTypes, Commitments, Connections, ContractChannels, NextChannelId,
+        NextClientId, NextConnectionId, QueryStore, WhitelistedRelayers, WhitelistedRelayersAdmin,
     },
     ContractError,
 };
@@ -633,9 +633,13 @@ pub fn execute(
                     client_id: msg.client_id,
                 },
             )?;
-            let encoded = U256::from(status as u32);
-            deps.storage
-                .write::<ClientStatuses>(&msg.client_id, &encoded);
+            let key = ClientStatusPath {
+                client_id: msg.client_id,
+            }
+            .key();
+            let value = U256::from(status as u32).to_le_bytes();
+            let value_h256 = H256::new(value);
+            store_commit(deps, &key, &value_h256);
             Ok(Response::default())
         }
     }
@@ -2201,10 +2205,12 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> Result<Binary, ContractErr
             Ok(to_json_binary(&status)?)
         }
         QueryMsg::GetCommittedStatus { client_id } => {
-            let val = deps.storage.read::<ClientStatuses>(&client_id)?;
-            let raw = val.to_le_bytes()[0];
-            let status = Status::try_from(raw)
-                .map_err(|_| ContractError::InvalidClientStatusValue { value: raw as u32 })?;
+            let commit = read_commit(deps, &ClientStatusPath { client_id }.key())
+                .ok_or(ContractError::CommittedClientStatusNotFound { client_id })?;
+            let commit_bytes: [u8; 32] = *commit.get();
+            let u256 = U256::from_le_bytes(commit_bytes);
+            let status = Status::try_from(u256)
+                .map_err(|err| ContractError::InvalidClientStatusValue { value: err.value })?;
             Ok(to_json_binary(&status)?)
         }
         QueryMsg::GetChannels { contract } => {
--- a/cosmwasm/ibc-union/core/src/lib.rs
+++ b/cosmwasm/ibc-union/core/src/lib.rs
@@ -159,10 +159,15 @@ pub enum ContractError {
     #[error("sender is not the relayer admin")]
     OnlyRelayerAdmin,
     #[error(
+        "{} committed client status not found for client id {client_id}",
+        ContractErrorKind::from(self)
+    )]
+    CommittedClientStatusNotFound { client_id: ClientId },
+    #[error(
         "{} invalid committed status value: {value}",
         ContractErrorKind::from(self)
     )]
-    InvalidClientStatusValue { value: u32 },
+    InvalidClientStatusValue { value: u8 },
 }
 
 impl ContractErrorKind {
--- a/cosmwasm/ibc-union/core/src/state.rs
+++ b/cosmwasm/ibc-union/core/src/state.rs
@@ -5,7 +5,7 @@ use depolama::{value::ValueCodecViaEncoding, KeyCodec, Prefix, Store, ValueCodec
 use ibc_union_spec::{Channel, ChannelId, ClientId, Connection, ConnectionId};
 use unionlabs::{
     encoding::Bincode,
-    primitives::{ByteArrayExt, Bytes, H256, U256},
+    primitives::{ByteArrayExt, Bytes, H256},
 };
 
 macro_rules! id_key {
@@ -173,24 +173,6 @@ impl Store for ClientStates {
 }
 id_key!(ClientStates);
 bytes_value!(ClientStates);
-
-pub enum ClientStatuses {}
-impl Store for ClientStatuses {
-    const PREFIX: Prefix = Prefix::new(b"client_status");
-
-    type Key = ClientId;
-    type Value = U256;
-}
-id_key!(ClientStatuses);
-impl ValueCodec<U256> for ClientStatuses {
-    fn encode_value(value: &U256) -> Bytes {
-        value.to_le_bytes().into()
-    }
-
-    fn decode_value(raw: &Bytes) -> StdResult<U256> {
-        read_fixed_bytes(raw).map(U256::from_le_bytes)
-    }
-}
 
 pub enum ClientConsensusStates {}
 impl Store for ClientConsensusStates {
--- a/lib/ibc-union-spec/src/path.rs
+++ b/lib/ibc-union-spec/src/path.rs
@@ -7,7 +7,7 @@ use voyager_primitives::IbcStorePathKey;
 use crate::Packet;
 use crate::{
     types::{ChannelId, ClientId, ConnectionId},
-    Channel, Connection, IbcUnion,
+    Channel, Connection, IbcUnion, Status,
 };
 
 pub const IBC_UNION_COSMWASM_COMMITMENT_PREFIX: [u8; 1] = [0x00];
@@ -34,6 +34,7 @@ pub const CONNECTIONS: U256 = U256::from_limbs([2, 0, 0, 0]);
 pub const CHANNELS: U256 = U256::from_limbs([3, 0, 0, 0]);
 pub const PACKETS: U256 = U256::from_limbs([4, 0, 0, 0]);
 pub const PACKET_ACKS: U256 = U256::from_limbs([5, 0, 0, 0]);
+pub const CLIENT_STATUS: U256 = U256::from_limbs([6, 0, 0, 0]);
 
 #[cfg(feature = "ethabi")]
 #[must_use]
@@ -60,6 +61,7 @@ pub enum StorePath {
     Channel(ChannelPath),
     BatchReceipts(BatchReceiptsPath),
     BatchPackets(BatchPacketsPath),
+    ClientStatus(ClientStatusPath),
 }
 
 impl StorePath {
@@ -72,6 +74,7 @@ impl StorePath {
             StorePath::Channel(path) => path.key(),
             StorePath::BatchReceipts(path) => path.key(),
             StorePath::BatchPackets(path) => path.key(),
+            StorePath::ClientStatus(path) => path.key(),
         }
     }
 }
@@ -264,4 +267,31 @@ impl IbcStorePathKey for BatchPacketsPath {
     type Spec = IbcUnion;
 
     type Value = H256;
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+#[cfg_attr(
+    feature = "serde",
+    derive(serde::Serialize, serde::Deserialize),
+    serde(rename_all = "snake_case", deny_unknown_fields)
+)]
+pub struct ClientStatusPath {
+    pub client_id: ClientId,
+}
+
+impl ClientStatusPath {
+    #[must_use]
+    pub fn key(&self) -> H256 {
+        Keccak256::new()
+            .chain_update(CLIENT_STATUS.to_be_bytes())
+            .chain_update(U256::from(self.client_id.get()).to_be_bytes())
+            .finalize()
+            .into()
+    }
+}
+
+impl IbcStorePathKey for ClientStatusPath {
+    type Spec = IbcUnion;
+    type Value = Status;
 }
--- a/lib/ibc-union-spec/src/types.rs
+++ b/lib/ibc-union-spec/src/types.rs
@@ -65,7 +65,6 @@ macro_rules! id {
                 id.raw().into()
             }
         }
-
         impl From<NonZeroU32> for $T {
             fn from(id: NonZeroU32) -> Self {
                 <$T>::new(id)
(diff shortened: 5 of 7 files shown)


lib/ibc-union-spec/src/path.rs
```diff
@@ -265,3 +268,31 @@ impl IbcStorePathKey for BatchPacketsPath {
...
+            .chain_update(CLIENT_STATUS.to_be_bytes())
+            .chain_update(U256::from(self.client_id.get()).to_be_bytes())
+            .finalize()
+            .into()
+    }
+}
+
+impl IbcStorePathKey for ClientStatusPath {
+    type Spec = IbcUnion;
+    type Value = u8; // repr(u8) status stored as raw byte
```
benluelo: we should move the status enum into this crate

cosmwasm/ibc-union/core/msg/src/lightclient.rs
```diff
@@ -6,10 +6,11 @@ use unionlabs_primitives::Bytes;
 #[derive(Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize)]
 #[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
 #[serde(deny_unknown_fields, rename_all = "snake_case")]
+#[repr(u8)]
 pub enum Status {
-    Active,
-    Expired,
-    Frozen,
+    Active = 0,
```
benluelo: this needs to be non-zero (i.e. starting at 1) such that it can be reliably differentiated from an empty storage value

benluelo (commented):
nice work! this also needs to be implemented in the solidity ibc stack for feature parity.

cosmwasm/ibc-union/core/msg/src/msg.rs
```diff
@@ -50,6 +50,14 @@ pub enum ExecuteMsg {
...
     WriteAcknowledgement(MsgWriteAcknowledgement),
     MigrateState(MsgMigrateState),
+    CommitClientStatus(MsgCommitClientStatus),
+}
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct MsgCommitClientStatus {
+    pub client_id: ClientId,
+    pub height: Option<u64>,
```
benluelo: this field is unused?

benluelo:
also please verify your commits as per our [contributing guide](https://github.com/unionlabs/union/blob/0763bf05b5ef88d04bd1ee4e09a75b1c88e1b63b/CONTRIBUTING.md#opening-the-pr).

cosmwasm/ibc-union/core/src/contract.rs
```diff
@@ -2171,6 +2194,10 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> Result<Binary, ContractErr
             )?;
             Ok(to_json_binary(&status)?)
         }
+        QueryMsg::GetCommittedStatus { client_id, .. } => {
```
benluelo: if the field is unused, remove it from the msg

lib/ibc-union-spec/Cargo.toml
```diff
@@ -28,7 +28,7 @@ voyager-primitives = { workspace = true }
 ibc-union-spec = { workspace = true, features = ["ethabi", "schemars", "serde"] }
 
 [features]
-default = []
+default = ["serde"]
```
benluelo: remove, do not add default features

lib/ibc-union-spec/src/path.rs
```diff
@@ -265,3 +269,41 @@ impl IbcStorePathKey for BatchPacketsPath {
...
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+#[serde(deny_unknown_fields, rename_all = "snake_case")]
+#[repr(u8)]
+pub enum Status {
+    Active = 1,
+    Expired = 2,
+    Frozen = 3,
+}
+
+/// Represents the path to a client's committed status (Active, Expired, etc.)
```
benluelo:
```suggestion
/// Represents the path to a client's committed [`Status`].
```

lib/ibc-union-spec/src/path.rs
benluelo: use `Status` here

voyager/modules/state/ethereum/src/main.rs
```diff
@@ -566,6 +566,23 @@ impl Module {
...
+
+    #[instrument(skip_all,fields(chain_id = %self.chain_id, %height, %client_id))]
+    async fn query_client_status(
+        &self,
+        height: Height,
+        client_id: ClientId,
+    ) -> RpcResult<Option<u8>> {
+        let status = self
+            .query_smart::<_, u8>(
+                &ibc_union_msg::query::QueryMsg::GetStatus { client_id },
```
benluelo: this is the wrong query; you need to use `GetCommittedStatus`

benluelo (changes requested):

cosmwasm/ibc-union/core/msg/src/query.rs
```diff
@@ -16,8 +16,10 @@ pub enum QueryMsg {
     GetClientState { client_id: ClientId },
     #[cfg_attr(feature = "cw-orch-interface", returns(unionlabs_primitives::Bytes))]
     GetConsensusState { client_id: ClientId, height: u64 },
-    #[cfg_attr(feature = "cw-orch-interface", returns(crate::lightclient::Status))]
+    #[cfg_attr(feature = "cw-orch-interface", returns(ibc_union_spec::path::Status))]
     GetStatus { client_id: ClientId },
+    #[cfg_attr(feature = "cw-orch-interface", returns(ibc_union_spec::path::Status))]
+    GetCommittedStatus { client_id: ClientId },
```
aeryz, line 22: `GetStatusCommitment` is more correct because we are not getting the `Status` that are committed. We are getting the commitment of the `Status` which is a hash of the encoded `Status`.

aeryz (dismissed):
Let's do the rename in the following PR.

lib/ibc-union-spec/src/path.rs
```diff
@@ -265,3 +268,45 @@ impl IbcStorePathKey for BatchPacketsPath {
...
+
+#[derive(Debug, Clone, PartialEq)]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(
+    feature = "serde",
+    serde(rename_all = "snake_case", deny_unknown_fields)
+)]
+#[repr(u8)]
+pub enum Status {
```
benluelo: this type should be under ibc_union_spec::types, and reexported at the crate root with the rest of the types.

cosmwasm/ibc-union/core/src/contract.rs
```diff
@@ -596,6 +597,28 @@ pub fn execute(
...
+
+            let status_bytes = (status as u8).to_le_bytes();
+
+            store_commit(
+                deps.branch(),
+                &ClientStatusPath {
+                    client_id: msg.client_id,
+                }
+                .key(),
+                &commit(status_bytes),
```
benluelo: store the bytes directly, as a `U256` value

voyager/modules/state/ethereum/src/main.rs
```diff
@@ -566,6 +566,15 @@ impl Module {
             None::<()>,
         ))
     }
+
+    #[instrument(skip_all,fields(chain_id = %self.chain_id, %height, %client_id))]
+    async fn query_client_status(
+        &self,
+        height: Height,
+        client_id: ClientId,
+    ) -> RpcResult<Option<u8>> {
```
benluelo: this should return `Option<Status>`

voyager/modules/state/cosmos-sdk-union/src/main.rs
```diff
@@ -367,6 +367,29 @@ impl Module {
...
+            chain_id = %self.chain_id,
+            %height,
+            %client_id
+        )
+    )]
+    async fn query_client_status(
+        &self,
+        height: Height,
+        client_id: ClientId,
+    ) -> RpcResult<Option<u8>> {
```
benluelo: `Option<Status>`

cosmwasm/ibc-union/lightclient/berachain/src/client.rs
```diff
@@ -5,7 +5,7 @@ use ethereum_light_client_types::StorageProof;
 use ibc_union_light_client::{
     spec::Timestamp, ClientCreationResult, IbcClient, IbcClientCtx, IbcClientError, StateUpdate,
 };
-use ibc_union_msg::lightclient::Status;
+use ibc_union_spec::path::Status;
```
benluelo: no need to add this as a dependency, `ibc_union_light_client` reexports `ibc_union_spec` under `ibc_union_light_client::spec` (as you can see in the import above this one)

cosmwasm/ibc-union/lightclient/berachain/Cargo.toml
```diff
@@ -28,6 +28,7 @@ evm-storage-verifier         = { workspace = true }
 frissitheto                  = { workspace = true }
 ibc-union-light-client       = { workspace = true }
 ibc-union-msg                = { workspace = true }
+ibc-union-spec               = { workspace = true }
```
benluelo: remove

cosmwasm/ibc-union/lightclient/base/src/client.rs
```diff
@@ -5,7 +5,7 @@ use ethereum_light_client_types::StorageProof;
 use ibc_union_light_client::{
     spec::Timestamp, ClientCreationResult, IbcClient, IbcClientCtx, IbcClientError, StateUpdate,
 };
-use ibc_union_msg::lightclient::Status;
+use ibc_union_spec::path::Status;
```
benluelo: same comment here about the reexport

cosmwasm/ibc-union/lightclient/base/Cargo.toml
```diff
@@ -28,6 +28,7 @@ evm-storage-verifier        = { workspace = true }
 frissitheto                 = { workspace = true }
 ibc-union-light-client      = { workspace = true }
 ibc-union-msg               = { workspace = true }
+ibc-union-spec              = { workspace = true }
```
benluelo: remove

cosmwasm/ibc-union/lightclient/arbitrum/src/client.rs
```diff
@@ -5,7 +5,7 @@ use ethereum_light_client_types::StorageProof;
 use ibc_union_light_client::{
     spec::Timestamp, ClientCreationResult, IbcClient, IbcClientCtx, IbcClientError, StateUpdate,
 };
-use ibc_union_msg::lightclient::Status;
+use ibc_union_spec::path::Status;
```
benluelo: use reexport

cosmwasm/ibc-union/lightclient/arbitrum/Cargo.toml
```diff
@@ -24,6 +24,7 @@ ethereum-light-client-types = { workspace = true }
 frissitheto                 = { workspace = true }
 ibc-union-light-client      = { workspace = true }
 ibc-union-msg               = { workspace = true }
+ibc-union-spec              = { workspace = true }
```
benluelo: remove

voyager/modules/state/cosmos-sdk-union/src/main.rs
```diff
@@ -367,6 +367,29 @@ impl Module {
...
+
+    #[instrument(
+        skip_all,
+        fields(
+            chain_id = %self.chain_id,
+            %height,
+            %client_id
+        )
+    )]
+    async fn query_client_status(
```
benluelo: rename this to `query_committed_client_status` (in the evm impl as well)

benluelo (changes requested):

cosmwasm/ibc-union/core/src/contract.rs
```diff
@@ -624,6 +626,24 @@ pub fn execute(
...
+                    client_id: msg.client_id,
+                },
+            )?;
+            store_commit(
+                deps.branch(),
+                &ClientStatusPath {
+                    client_id: msg.client_id,
+                }
+                .key(),
+                &commit(U256::from(status as u32).to_le_bytes()),
```
benluelo: don't store the commit, but the u256 value directly

cosmwasm/ibc-union/core/msg/src/query.rs
```diff
@@ -16,8 +16,10 @@ pub enum QueryMsg {
     GetClientState { client_id: ClientId },
     #[cfg_attr(feature = "cw-orch-interface", returns(unionlabs_primitives::Bytes))]
     GetConsensusState { client_id: ClientId, height: u64 },
-    #[cfg_attr(feature = "cw-orch-interface", returns(crate::lightclient::Status))]
+    #[cfg_attr(feature = "cw-orch-interface", returns(ibc_union_spec::Status))]
     GetStatus { client_id: ClientId },
+    #[cfg_attr(feature = "cw-orch-interface", returns(Option<H256>))]
```
benluelo: return status here

cosmwasm/ibc-union/core/src/contract.rs
```diff
@@ -624,6 +625,19 @@ pub fn execute(
...
+        ExecuteMsg::CommitClientStatus(msg) => {
+            let status = query_light_client::<Status>(
+                deps.as_ref(),
+                client_impl(deps.as_ref(), msg.client_id)?,
+                LightClientQuery::GetStatus {
+                    client_id: msg.client_id,
+                },
+            )?;
+            let encoded = U256::from(status as u32);
+            deps.storage
```
benluelo: this needs to be written to the commitment store, *not* the regular storage. it needs to be committed as the un-hashed value. use `store_commit`.

cosmwasm/ibc-union/core/src/contract.rs
```diff
@@ -2186,6 +2200,13 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> Result<Binary, ContractErr
             )?;
             Ok(to_json_binary(&status)?)
         }
+        QueryMsg::GetCommittedStatus { client_id } => {
+            let val = deps.storage.read::<ClientStatuses>(&client_id)?;
```
benluelo: same comment as above, use `read_commit` here.

cosmwasm/ibc-union/core/src/state.rs
```diff
@@ -174,6 +174,24 @@ impl Store for ClientStates {
...
+impl ValueCodec<U256> for ClientStatuses {
+    fn encode_value(value: &U256) -> Bytes {
+        value.to_le_bytes().into()
+    }
+
+    fn decode_value(raw: &Bytes) -> StdResult<U256> {
+        read_fixed_bytes(raw).map(U256::from_le_bytes)
+    }
+}
+
```
benluelo: remove this store

lib/ibc-union-spec/src/types.rs
```diff
@@ -111,3 +112,16 @@ impl fmt::Display for Status {
...
+
+    fn try_from(value: u8) -> Result<Self, Self::Error> {
+        match value {
+            1 => Ok(Status::Active),
+            2 => Ok(Status::Expired),
+            3 => Ok(Status::Frozen),
+            _ => Err("Invalid status value"),
+        }
+    }
+}
```
benluelo: use `UnknownEnumVariant<u8>` here

lib/ibc-union-spec/src/types.rs
```diff
@@ -111,3 +112,16 @@ impl fmt::Display for Status {
         })
     }
 }
+
+impl TryFrom<u8> for Status {
```
benluelo, line 120: also add a `TryFrom<U256>` impl

cosmwasm/ibc-union/core/src/lib.rs
```diff
@@ -158,6 +158,11 @@ pub enum ContractError {
     OnlyWhitelistedRelayer,
     #[error("sender is not the relayer admin")]
     OnlyRelayerAdmin,
+    #[error(
+        "{} invalid committed status value: {value}",
+        ContractErrorKind::from(self)
+    )]
+    InvalidClientStatusValue { value: u32 },
```
benluelo: why use u32?

cosmwasm/ibc-union/core/src/contract.rs
```diff
@@ -2186,6 +2200,13 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> Result<Binary, ContractErr
...
+            let raw = val.to_le_bytes()[0];
+            let status = Status::try_from(raw)
+                .map_err(|_| ContractError::InvalidClientStatusValue { value: raw as u32 })?;
```
benluelo: don't arbitrarily cast to an integer width. you need to consider all possible values this may take. additionally, since we control this error type, we don't need to cast to a specific integer width - we can just change the field type (as i mentioned in another comment).

benluelo (changes requested):

lib/ibc-union-spec/src/types.rs
```diff
@@ -113,15 +112,29 @@ impl fmt::Display for Status {
...
+            _ => Err(UnknownEnumVariant { value }),
         }
     }
 }
+
+impl TryFrom<U256> for Status {
+    type Error = UnknownEnumVariant;
+
+    fn try_from(value: U256) -> Result<Self, Self::Error> {
+        let byte = value.to_le_bytes()[0];
```
benluelo, line 137: this drops the majority of the u256. implement this correctly

lib/ibc-union-spec/src/types.rs
```diff
@@ -111,3 +111,30 @@ impl fmt::Display for Status {
         })
     }
 }
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct UnknownEnumVariant {
+    pub value: u8,
+}
```
benluelo, line 118: don't redefine this, reuse it from `unionlabs`

PR #4860, opened 2025-07-20
feat(voyager): latest height/timestamp caching

benluelo:


lib/voyager-core/src/cache.rs
```diff
@@ -60,6 +66,26 @@ impl Cache {
...
+
+            // 100 should be enough but would ideally be a passed in value
+            latest_timestamp_cache: moka::future::CacheBuilder::new(100)
+                .time_to_live(Duration::from_secs(1000 * 365 * 24 * 60 * 60))
+                .time_to_idle(Duration::from_secs(1000 * 365 * 24 * 60 * 60))
+                .eviction_policy(EvictionPolicy::lru())
+                .build(),
+            latest_timestamp_metric: opentelemetry::global::meter("voyager")
+                .u64_gauge("chain.latest_timestamp")
+                .build(),
```
benluelo, line 88: this isn't the "best" way to do this, but it works for now

hussein-aitlahcen (approved):

PR #4875, opened 2025-07-21
finalize base and upgrade bob light client/modules

hussein-aitlahcen:


voyager/plugins/client-update/bob/src/main.rs
```diff
@@ -407,13 +449,16 @@ impl Module {
...
+        let ClientState::V2(bob_client_state) = voy_client
             .decode_client_state::<IbcUnion, ClientState>(
                 bob_client_state_info.client_type.clone(),
                 bob_client_state_info.ibc_interface,
                 raw_bob_client_state,
             )
-            .await?;
+            .await?
+        else {
+            panic!("expected client state v2")
```
benluelo: nit: would be good to return FATAL_ERROR here

PR #4881, opened 2025-07-22
fix: Incorrect field element initialization in Update zk_verifier.go

HarleyQeen:
This PR fixes a critical bug in the inputsHash function where a truncated 31-byte slice (hash[1:]) was passed to fr.Element.SetBytes. This violates the expected 32-byte input and may lead to:

Incorrect field element interpretation.

Undetected ZKP verification failures.

Security issues due to reduced entropy or misaligned hash mapping. This change ensures that the full SHA-256 output (32 bytes) is passed as input to the field element constructor, preserving correctness and cryptographic soundness.

Impact:
Fixes a silent but critical bug that could cause ZKP verification failures or incorrect behavior depending on the input.

Does not affect external API.

Maintains full compatibility with properly generated proofs.

Notes:
This was previously “working” likely due to coincidental inputs or incomplete test coverage.

This fix aligns with the expected usage of fr.Element.SetBytes() in gnark-crypto.

### DIFF
--- a/11-cometbls/zk_verifier.go
+++ b/11-cometbls/zk_verifier.go
@@ -212,6 +212,6 @@ func inputsHash(header ProverLightHeader, trustedValidatorsHash []byte) fr.Eleme
 	hash := sha256.Sum256(buff)
 
 	var e fr.Element
-	e.SetBytes(hash[1:])
+	e.SetBytes(hash[:])
 	return e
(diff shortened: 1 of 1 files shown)


aeryz:
This is done intentionally to fit the output to the BN254 curve and has no major security impact.

PR #4937, opened 2025-08-04
feat(cosmwasm): introduce escrow vault for fungibility layer

hussein-aitlahcen:
A CosmWasm contract that acts as a solver in the zkgm protocol, enabling fungible token transfers across IBC channels by managing escrowed assets.

  Key Features

  - Dual Token Support: Handles both native Cosmos tokens (via BankMsg) and CW20 tokens
  - Fungible Lane Configuration: Maps (channel_id, base_token) pairs to counterparty beneficiaries for cross-chain settlement
  - Intent Whitelisting: Controls which packet hashes can be filled before finality
  - Solver Interface: Implements DoSolve to release escrowed tokens and return beneficiary addresses for acknowledgments

  How It Works

  1. Admin configures fungible lanes linking channels/tokens to counterparty beneficiaries
  2. When zkgm calls DoSolve, the contract validates the order and releases escrowed tokens
  3. Relayers receive fees (base_amount - quote_amount)
  4. Returns the counterparty beneficiary address for source chain repayment (symmetric vault)

  Security

  - Access controlled: Only zkgm can call DoSolve, only admin can configure lanes
  - Validates base_amount ≥ quote_amount to prevent underflows
  - Intent packets must be whitelisted to prevent unauthorized pre-finality fills

### DIFF
--- a/cosmwasm/cw-escrow-vault/src/contract.rs
+++ b/cosmwasm/cw-escrow-vault/src/contract.rs
@@ -1,0 +1,190 @@
+use cosmwasm_schema::cw_serde;
+use cosmwasm_std::{
+    entry_point, to_json_binary, wasm_execute, BankMsg, Binary, Coin, CosmosMsg, Deps, DepsMut,
+    Env, MessageInfo, Response, StdResult, Uint128,
+};
+use cw20::Cw20ExecuteMsg;
+use depolama::StorageExt;
+use frissitheto::UpgradeMsg;
+use ibc_union_spec::path::commit_packets;
+use unionlabs::primitives::U256;
+
+use crate::{
+    error::Error,
+    msg::{ExecuteMsg, InstantiateMsg, QueryMsg},
+    state::{Admin, FungibleCounterparty, FungibleLane, IntentWhitelist, Zkgm},
+};
+
+#[entry_point]
+pub fn instantiate(_: DepsMut, _: Env, _: MessageInfo, _: ()) -> StdResult<Response> {
+    panic!("this contract cannot be instantiated directly, but must be migrated from an existing instantiated contract.");
+}
+
+#[cw_serde]
+pub struct MigrateMsg {}
+
+#[entry_point]
+pub fn migrate(
+    deps: DepsMut,
+    _: Env,
+    msg: UpgradeMsg<InstantiateMsg, MigrateMsg>,
+) -> Result<Response, Error> {
+    msg.run(
+        deps,
+        |deps, msg| {
+            deps.storage.write_item::<Admin>(&msg.admin);
+            Ok((Response::new(), None))
+        },
+        |_, _, _| Ok((Response::default(), None)),
+    )
+}
+
+fn ensure_zkgm(deps: Deps, info: &MessageInfo) -> Result<(), Error> {
+    let admin = deps.storage.read_item::<Zkgm>()?;
+    if info.sender != admin {
+        return Err(Error::OnlyZkgm);
+    }
+    Ok(())
+}
+
+fn ensure_admin(deps: Deps, info: &MessageInfo) -> Result<(), Error> {
+    let admin = deps.storage.read_item::<Admin>()?;
+    if info.sender != admin {
+        return Err(Error::OnlyAdmin);
+    }
+    Ok(())
+}
+
+#[entry_point]
+pub fn execute(
+    deps: DepsMut,
+    _env: Env,
+    info: MessageInfo,
+    msg: ExecuteMsg,
+) -> Result<Response, Error> {
+    match msg {
+        ExecuteMsg::WhitelistIntents { hashes_whitelist } => {
+            ensure_admin(deps.as_ref(), &info)?;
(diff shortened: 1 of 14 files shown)


cosmwasm/cw-escrow-vault/src/contract.rs
```diff
@@ -0,0 +1,183 @@
...
+                        );
+                    }
+                }
+                Ok(())
+            };
+
+            let fee = order
+                .base_amount
+                .checked_sub(order.quote_amount)
+                .expect("impossible");
```
benluelo: perhaps I'm missing some context but why is this impossible?

cosmwasm/cw-escrow-vault/src/state.rs
```diff
@@ -0,0 +1,109 @@
...
+    type Key = ();
+    type Value = Addr;
+}
+
+impl ValueCodec<Addr> for Admin {
+    fn encode_value(value: &Addr) -> Bytes {
+        value.as_bytes().into()
+    }
+
+    fn decode_value(raw: &Bytes) -> StdResult<Addr> {
```
benluelo, line 56: Todo for me: add a default addr value codec impl somehow in depolama

cosmwasm/cw-escrow-vault/src/msg.rs
```diff
@@ -0,0 +1,38 @@
...
+#[cw_serde]
+pub struct InstantiateMsg {
+    pub zkgm: Addr,
+    pub admin: Addr,
+}
+
+#[cw_serde]
+pub enum ExecuteMsg {
+    WhitelistIntents {
+        hashes_whitelist: Vec<(H256, bool)>,
```
benluelo, line 17: this allows to blacklist packets as well then?

cosmwasm/cw-escrow-vault/src/state.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    fn decode_key(raw: &Bytes) -> StdResult<H256> {
+        Ok(H256::new(raw.as_ref().try_into().expect("impossible")))
+    }
+}
+impl ValueCodecViaEncoding for IntentWhitelist {
+    type Encoding = Bincode;
+}
+
+pub type BaseToken = Bytes;
```
benluelo: either newtype wrapped or don't use an alias

cosmwasm/cw-escrow-vault/src/contract.rs
hussein-aitlahcen: we check that base covers quote in term of amount just before

cosmwasm/cw-escrow-vault/src/msg.rs
hussein-aitlahcen: yes, it overwritable so you can whitelist and if you change your mind overwrite to blacklist which is like not whitelisting at all

cosmwasm/cw-escrow-vault/src/contract.rs
benluelo: ah ok i missed that line. maybe good to do the comparison as the checked_sub call then? then we can match on the result, instead of doing the check twice?

cosmwasm/cw-escrow-vault/src/contract.rs
```diff
@@ -0,0 +1,185 @@
...
+                    .read::<IntentWhitelist>(&commit_packets(&[packet.clone()]))
+                    .unwrap_or(false);
+                if !whitelisted {
+                    return Err(Error::IntentMustBeWhitelisted);
+                }
+            }
+
+            let fungible_lane = deps
+                .storage
+                .read::<FungibleCounterparty>(&(
```
benluelo: use maybe_read instead which returns an option on nonexistence

cosmwasm/cw-escrow-vault/src/state.rs
```diff
@@ -0,0 +1,108 @@
...
+    type Key = H256;
+    type Value = bool;
+}
+impl KeyCodec<H256> for IntentWhitelist {
+    fn encode_key(key: &H256) -> Bytes {
+        key.into_bytes()
+    }
+
+    fn decode_key(raw: &Bytes) -> StdResult<H256> {
+        Ok(H256::new(raw.as_ref().try_into().expect("impossible")))
```
benluelo: this is fine for now but decoding shouldn't panic

cosmwasm/cw-escrow-vault/src/state.rs
```diff
@@ -0,0 +1,108 @@
...
+
+pub enum FungibleCounterparty {}
+impl Store for FungibleCounterparty {
+    const PREFIX: Prefix = Prefix::new(b"fungible_counterparty");
+
+    type Key = (Uint256, ChannelId, Bytes);
+    type Value = FungibleLane;
+}
+
+impl KeyCodec<(Uint256, ChannelId, Bytes)> for FungibleCounterparty {
```
benluelo: if you use a U256 from unionlabs primitives you can use bincode encoding here instead of manually doing it

PR #4938, opened 2025-08-04
feat(zkgm): secure wrapped token event

hussein-aitlahcen:


cosmwasm/osmosis-tokenfactory-token-minter/src/contract.rs
```diff
@@ -234,30 +265,44 @@ fn wrapped_create_denom(
...
+                value: MsgSetDenomMetadata {
+                    sender: env.contract.address.to_string(),
+                    metadata: Some(new_proto_metadata(denom.clone(), metadata)?),
+                }
+                .encode_to_vec()
+                .into(),
+            },
+        ])
+        .add_event(
+            Event::new("new_secure_wrapped_token")
```
benluelo: we should export these as consts from the lib crate

benluelo (approved):

PR #4954, opened 2025-08-07
Update @cosmjs__amino@0.33.1.patch

serkanmax:
Add support for Bn254 public key type with null-safe checks

- Introduced `isBn254Pubkey()` function to support Tendermint Bn254 public keys.
- Added `bn254` to `pubkeyType` map.
- Used optional chaining (`pubkey?.type`) in all type check functions to avoid potential TypeErrors when input is null or undefined.

(just a novice suggestion from a novice user)


benluelo:
this patch file is very fragile, please do not touch it.

PR #4990, opened 2025-08-13
feat(voyager): individual packet index plugin

benluelo:


quietBlockchain:
Great addition! Having the Packet Index plugin and the ability to query packets/acks by hash will be super helpful for debugging and monitoring. Excited to see how this improves cross-chain observability. Thanks for shipping this!

ayyubs01:
Oxba5ed

PR #5032, opened 2025-08-25
fix(app): reduce redundant logging

ehegnes:


deployments/editions.json
```diff
@@ -41,17 +41,11 @@
       "bob.60808": {
         "environment": "production"
       },
-      "bsc.56": {
```
qlp, line 44: intentionally?

cor (approved):

PR #5033, opened 2025-08-25
Cosmwasm testnet zkgm migration v2

hussein-aitlahcen:


DevLabsss:
only i know and a few people here know how hard dev works

zkgm dev

PR #5048, opened 2025-08-27
feat(zkgm): solver based staking instructions

hussein-aitlahcen:


cosmwasm/ucs03-zkgm-api/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+
+[dependencies]
+alloy-primitives  = { workspace = true }
+alloy-sol-types   = { workspace = true }
+cosmwasm-schema   = { workspace = true }
+cosmwasm-std      = { workspace = true, features = ["std"] }
+enumorph          = { workspace = true }
+ibc-union-spec    = { workspace = true, features = ["schemars", "serde"] }
+token-factory-api = { workspace = true }
+unionlabs         = { workspace = true, default-features = false }
```
benluelo: default-features = false here doesn't do anything, and i think cargo may complain about it

docs/src/content/docs/ucs/03.mdx
```diff
@@ -992,6 +991,10 @@ The stake instruction allows cross-chain staking of governance tokens with valid
 3. A staking position NFT is minted to the beneficiary
 4. The NFT represents ownership of the staked position and can be used to manage it
 
+:::note
+The `path` field in the packet must be zero for staking operations. The staked token must match the governance token configured for the channel.
```
benluelo, line 995: why do we check against the configured gov token, rather than just use the configured value?

hussein-aitlahcen: because the packet is immutable

PR #5093, opened 2025-09-02
Add guards for zero gameCount to prevent U256 underflow

MozirDmitriy:
Introduce explicit zero-count checks in voyager client-update plugins and use saturating subtraction in base-client and bob-client when deriving the latest game index from gameCount. This prevents U256 underflow when no dispute games exist at a given L1 block and returns a clear error to callers instead of panicking.


benluelo:
there have been significant changes with voyager plugin/module error handling; the work done here is no longer applicable. the changes can be reapplied but will need to be done with the new error types. closing this for now, please open a new PR if you want to pick this back up!

PR #5098, opened 2025-09-02
feat(uniond): increase cost of wasmd tx storage

PoisonPhang:


PoisonPhang:
closing in favor of different approach

PR #5113, opened 2025-09-03
chore(zkgm): allow empty/invalid mm address on ack

hussein-aitlahcen:


cosmwasm/ibc-union/app/ucs03-zkgm/src/contract.rs
```diff
@@ -972,23 +972,21 @@ fn acknowledge_fungible_asset_order_v2(
...
-                                amount: base_amount.into(),
-                                mint_to_address: market_maker,
-                            },
-                            minter,
-                            vec![],
-                        )?);
+                        if let Some(market_maker) = str::from_utf8(ack.market_maker.as_ref())
+                            .ok()
+                            .and_then(|x| deps.api.addr_validate(x).ok())
+                        {
```
benluelo, line 979: maybe emit an event if unable to parse?

PR #5127, opened 2025-09-05
fix: package set and entrypoint

famouswizard:
changed `packages` to an attribute set instead of `//` to avoid errors with derivations.
aso wrapped the entrypoint in `"${...}"` so Docker gets strings, not derivations.

**chngs**

* `packages = { mpc-coordinator = mpc-coordinator; mpc-client = mpc-client; mpc-client-image = ...; }`
* `entrypoint = [ "${pkgs.lib.getExe self'.packages.mpc-client}" ]`


benluelo:
there are no issues with the existing code, and the new code is worse in multiple ways.

PR #5217, opened 2025-10-03
Voyager sui plugin

aeryz:


sui/u/sources/u.move
```diff
@@ -0,0 +1,82 @@
...
+    public struct U has drop {}
+
+    fun init(witness: U, ctx: &mut TxContext) {
+        let (treasury_cap, metadata) =
+            coin::create_currency<U>(
+                witness,
+                6,
+                b"muno",
+                b"muno",
+                b"U",
```
benluelo: muno -> au

sui/ucs03_zkgm/Move.lock
benluelo, line 1: are move workspaces a thing? would be nicer if we can not have a .lock for every contract

voyager/modules/state/sui/src/main.rs
```diff
@@ -295,7 +295,7 @@ impl StateModuleServer<IbcUnion> for Module {
 
                 into_value(consensus_state_bytes)
             }
-            _ => todo!(),
+            what => panic!("WHAT: {what:?}"),
```
benluelo, line 298: agreed

voyager/plugins/sui-ibc-app/src/lib.rs
benluelo, line 1: remove? empty

sui/ucs03_zkgm/Move.lock
aeryz: why do you think we can have good things?

sui/u/sources/u.move
aeryz: hmmm lemme introduce u and muno separately then. in our devnet its still muno right?

benluelo: lets change devnet to U

PR #5246, opened 2025-10-17
Sui Escrow Vault + More Unit Tests

Caglankaan:


sui/ucs03_zkgm/sources/zkgm.move
```diff
@@ -946,6 +969,10 @@ module zkgm::zkgm {
             return (vector::empty(), E_INVALID_QUOTE_TOKEN)
         };
 
+        if(zkgm.test_mode){
+            return (vector::empty(), 0)
+        };
+        
         vault.solve<T>(
```
aeryz: if the `test_mode` is only needed in here, can't we make `vault::solve` work? iirc, it only requires a minimal setup where the fungible counterparty must be set and that's it.

sui/vault/Move.toml
```diff
@@ -1,20 +1,20 @@
...
 
 [addresses]
-# ibc   = "0xd5b50a4b934303c39baf1999ddf84cfcd65b4c6f73b2226bfea77987c21fbff8"
-# vault = "0x0"
+# ibc   = "_"
+# vault = "_"
 ibc   = "_"
 vault = "_"
 
-[dev-dependencies]
```
aeryz, line 17: are these addresses being uncommented causing any issues?

aeryz (approved):

PR #5254, opened 2025-10-21
Add development environment for sui.103

Caglankaan:


deployments/editions.json
```diff
@@ -37,6 +37,9 @@
       "sei.1329": {
         "environment": "production"
       },
+      "sui.103": {
```
qlp, line 40: that's not a valid chain-id?

thought `35834a8a` is mainnet and `4c78adac` is testnet

Caglankaan: i don't know i just take a look what sui testnet chain id is, it was 103 so i wrote sui.103

qlp: where did you find this?

Caglankaan: https://docs.expand.network/ids/chain-ids

qlp: I can't find this in the sui docs. we're expecting the value returned by sui_getChainIdentifier: https://docs.sui.io/sui-api-ref#sui_getchainidentifier

PR #5256, opened 2025-10-21
feat(lst): redelegations

benluelo:


cosmwasm/lst-staker/Cargo.toml
```diff
@@ -34,4 +34,5 @@ hex-literal = { workspace = true }
 
 [features]
 default  = []
+library  = []
```
benluelo, line 37: why?

tools/union-test/Cargo.toml
```diff
@@ -45,6 +45,7 @@ unionlabs                      = { workspace = true }
 voyager-sdk                    = { workspace = true }
 
 [dev-dependencies]
+lst-staker         = { workspace = true, features = ["library"] }
```
benluelo, line 48: ah, that's why

cosmwasm/lst-staker/src/lib.rs
```diff
@@ -64,100 +66,292 @@ pub fn execute(
...
+    // sitting in the contract after we unstake
+    let (rebase_response, total_pending_rewards) = rebase(deps, env)?;
+
+    Ok(Response::new()
+        .add_events(rebase_response.events)
+        .add_submessages(rebase_response.messages)
+        .add_event(
+            Event::new("unstake")
+                .add_attribute("total", amount.to_string())
+                .add_attribute("pending_rewards", total_pending_rewards),
```
benluelo: no need for this, doesn't need to be threaded through `rebase`

aeryz: YOU ADDED IT

PR #5270, opened 2025-10-24
feat(cosmwasm): init upgradable

benluelo:


cosmwasm/upgradable/src/lib.rs
```diff
@@ -0,0 +1,34 @@
...
+
+use cosmwasm_std::{DepsMut, Env, Event, MessageInfo, Response, WasmMsg, to_json_binary};
+use frissitheto::UpgradeMsg;
+
+use crate::{error::ContractError, msg::ExecuteMsg};
+
+pub mod error;
+pub mod msg;
+
+#[cfg_attr(not(feature = "library"), cosmwasm_std::entry_point)]
```
aeryz, line 13: this contract has no way of being a contract by it's own right? I think we should make it a proper library, with no `execute` but proper functions (eg `fn upgrade`)

benluelo: yeah we can do that. i want to rework our cosmwasm/ directory to have a lib/ as well, and i'll move some other libraries we have when i do this, but for now i'll keep it like this since i really don't want to rewrite this rn lol

aeryz: you owe me one "approve but no questions asked"

aeryz (approved):

PR #5300, opened 2025-11-10
attestation client

benluelo:


cosmwasm/lightclient/attested/src/execute.rs
```diff
@@ -0,0 +1,227 @@
...
+        return Err(Error::AlreadyAttested {
+            chain_id: attestation.chain_id.clone(),
+            height: attestation.height,
+            timestamp: attestation.timestamp,
+            key: attestation.key,
+            value,
+        });
+    }
+
+    ensure!(
```
aeryz, line 64: considering we might have gud amount of traffic, maybe we can change the api to be able to verify multiple attestations at once through `cosmwasm`'s batch verification api.

cosmwasm/ibc-union/lightclient/attested/src/msg.rs
```diff
@@ -1,25 +1,72 @@
...
+    #[serde(untagged)]
+    Restricted(RestrictedExecuteMsg),
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(deny_unknown_fields, rename_all = "snake_case")]
+pub enum RestrictedExecuteMsg {
+    /// Set a new quorum for the attestations to be considered valid.
+    ///
+    /// If the new quorum is larger than the currently configuured quorum, any existing attestations that have already hit the quorum will still be considered valid, but any current pending attestations will need to reach the new quorum in order to be confirmed.
```
aeryz:
```suggestion
    /// If the new quorum is larger than the currently configured quorum, any existing attestations that have already hit the quorum will still be considered valid, but any current pending attestations will need to reach the new quorum in order to be confirmed.
```

cosmwasm/ibc-union/lightclient/attested/src/msg.rs
```diff
@@ -1,25 +1,72 @@
...
+#[serde(deny_unknown_fields, rename_all = "snake_case")]
+pub enum QueryMsg {
+    /// Returns the currently configured quorum for `chain_id`.
+    Quorum { chain_id: String },
+    /// Returns the current attestation set for `chain_id`.
+    Attestors { chain_id: String },
+    /// Returns the value attested to under `key` at `height` on `chain_id`.
+    AttestedValue {
+        chain_id: String,
+        // #[serde(with = "serde_utils::string")]
```
aeryz: shouldn't we enable this?

cosmwasm/ibc-union/lightclient/attested/src/msg.rs
```diff
@@ -1,25 +1,72 @@
...
+        height: u64,
+        key: Bytes,
+    },
+    /// Returns the timestamp attested to at `height` on `chain_id`.
+    TimestampAtHeight {
+        chain_id: String,
+        // #[serde(with = "serde_utils::string")]
```
aeryz: this as well

cosmwasm/lightclient/attested/src/execute.rs
benluelo: i considered this, it would be a trivial addition down the line on the contract side but it will be a decent amount more work on the voyager side so i decided to just go with this MVP for now

cosmwasm/ibc-union/lightclient/attested/src/msg.rs
benluelo: yes, i just needed to add the dependency, put the comment so it wouldn't be forgotten lol

PR #5316, opened 2025-11-21
feat(devshell): configure zed to use devShell LSPs

cor:
Configures [Zed](https://zed.dev/) to use Union's devShell. Zed is an extremely fast IDE written in Rust with optional Helix keybindings and excellent AI Agent (like Claude Code) integration. It works over SSH so you can connect to your NixOS VM from macOS.

Screenshot 2025-11-21 at 15 24 31

### DIFF
--- a/.zed/settings.json
+++ b/.zed/settings.json
@@ -1,0 +1,49 @@
+// Zed settings to use Union's devShell
+{
+  "languages": {
+    "TypeScript": {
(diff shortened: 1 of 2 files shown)


.zed/settings.json
benluelo, line 1: should probably be .jsonc?

cor: Nope, zed said .json it is and inserted comments anyways

PR #5318, opened 2025-11-24
feat(private-transfer): introduce ZAsset

hussein-aitlahcen:


evm/contracts/ZAsset.sol
```diff
@@ -0,0 +1,957 @@
...
+            Counterparty storage counterparty =
+                $.counterparties[lightClient.clientId];
+            if (counterparty.tokenAddressKey == 0) {
+                revert OnlyKnownLightClients();
+            }
+            bytes memory consensusState = IBCStore($.ibcHandler).getClient(
+                lightClient.clientId
+            ).getConsensusState(lightClient.clientId, lightClient.height);
+            (, bytes32 stateRoot,) =
+                abi.decode(consensusState, (uint64, bytes32, bytes32));
```
benluelo: the consensus states are not all necessarily the same shape, we probably want to use the offsets here like we do in the state lens clients (configured per chain)

benluelo (commented):
we should probably make the loopback client configurable in the deployment script, since it won't work on chains without the eip enabled

hussein-aitlahcen:
> we should probably make the loopback client configurable in the deployment script, since it won't work on chains without the eip enabled

It would still work as it rely on the blockhash opcode (non eip) if it's a recent (>tip-256) block https://github.com/Vectorized/solady/blob/73f13dd1483707ef6b4d16cb0543570b7e1715a8/src/utils/BlockHashLib.sol#L51.

If you are out of the window it would revert, but that would be as if the window is "smaller" so I guess we are safe.

PR #5323, opened 2025-11-26
feat(voyager): custom error type for plugins and modules

benluelo:


voyager/modules/finality/ethereum/src/main.rs
```diff
@@ -111,52 +107,42 @@ impl Module {
...
-                |f| {
-                    (
-                        f.finalized_header.execution.block_number,
-                        f.finalized_header.execution.timestamp,
-                    )
-                },
-            ))
+        Ok(self.finality_update_cached().await?.fold(
+            |f| match f {},
+            |_| todo!(),
```
aeryz, line 119: this exists in the previous version as well but how does this work?

aeryz (approved):

voyager/modules/finality/ethereum/src/main.rs
benluelo: how does what work? the empty match?

PR #5328, opened 2025-12-06
feat(starknet): lc & voyager

benluelo:


voyager/plugins/event-source/starknet/src/main.rs
```diff
@@ -0,0 +1,1247 @@
...
+                        port_id,
+                        channel_id,
+                        counterparty_port_id,
+                        counterparty_channel_id,
+                        connection_id,
+                        counterparty_version,
+                    };
+
+                    dbg!(event, emitted_event);
+                }
```
benluelo, line 391: @qlp this is the general logic for parsing starknet events, i'm going to pull this out into a library and you can reconsume it in hubble

PoisonPhang (approved):

PR #5340, opened 2025-12-15
feat(cosmwasm): clean up deployer and fix lc upgrade script

benluelo:


cosmwasm/cosmwasm.nix
```diff
@@ -13,7 +13,7 @@ _: {
     let
       crane = mkCrane {
         root = ../.;
-        # gitRev = "1b172ff9247e5eedcddc6665c856b0bf4427035d";
+        # gitRev = "";
```
PoisonPhang, line 16: > Do it, kill him now.

We can delete this entirely, yeah?

benluelo: this is used to deploy on a commit when there are local changes, i just forgot to remove the commit from here the last time i deployed

PoisonPhang (approved):

PR #5374, opened 2026-01-09
feat(docs): deployments and voyager docs cleanup

benluelo:
also more lenient packet hash parsing in app urls


ts-sdk/src/schema/packet.ts
```diff
@@ -11,9 +13,63 @@ import { PortId } from "./port.js"
...
+        }),
+        encode: ({ value }) => value,
+      },
+    )),
+).pipe(
+  S.brand("LenientPacketHash"),
+)
+export type LenientPacketHash = typeof LenientPacketHash.Type
+
+export const PacketHashFromLenient = S.transform(
```
ehegnes, line 55: Given that this is fallible, this should be `S.transforOrFail` and the inner `S.decodeUnknownSync` converted to `S.decodeUnknown` such that it produces a `ParseResult`.

benluelo: it's actually infallible due to the LenientPacketHash schema

PR #5400, opened 2026-01-21
revive parlia lc

benluelo:


hussein-aitlahcen (approved):

Cargo.toml
```diff
@@ -218,10 +225,7 @@ members = [
   # "lib/aptos-verifier",
 
   "lib/reconnecting-jsonrpc-ws-client",
-  "lib/voyager-primitives",
```
aeryz, line 221: what happened to these not used anymore?

cosmwasm/core/src/tests.rs
```diff
@@ -199,3 +200,31 @@ fn display() {
         .unwrap()
     )
 }
+
+#[test]
+fn serde() {
+    let msg = RestrictedExecuteMsg::ChannelOpenTry(MsgChannelOpenTry {
```
aeryz: let's either assert the result or remove the test entirely if we already have serde tests for the untagged stuff.

cosmwasm/lightclient/proof-lens/src/client.rs
```diff
@@ -0,0 +1,207 @@
...
+use ibc_union_spec::{
+    Status, Timestamp,
+    path::{ConsensusStatePath, MembershipProofPath, NON_MEMBERSHIP_COMMITMENT_VALUE},
+};
+use proof_lens_light_client_types::{ClientState, ConsensusState, Header};
+use unionlabs::{encoding::Bincode, ethereum::keccak256};
+
+use crate::{errors::Error, raw_bytes::RawBytes};
+
+pub struct StateLensIcs23Ics23LightClient;
```
aeryz:
```suggestion
pub struct ProofLensIcs23Ics23LightClient;
```

cosmwasm/lightclient/proof-lens/src/raw_bytes.rs
```diff
@@ -0,0 +1,23 @@
...
+#[derive(Debug)]
+pub struct RawBytes(pub Bytes);
+
+impl Encode<Bincode> for RawBytes {
+    fn encode(self) -> Vec<u8> {
+        self.0.into()
+    }
+}
+
+impl Decode<Bincode> for RawBytes {
```
aeryz, line 17: hmmmm this will create an invalid bincode (de)serialization when used within a type right? since you dunno how many bytes to parse.

evm/contracts/core/24-host/IBCCommitment.sol
```diff
@@ -7,6 +7,11 @@ library IBCCommitment {
     uint256 public constant CHANNELS = 0x03;
     uint256 public constant PACKETS = 0x04;
     uint256 public constant PACKET_ACKS = 0x05;
+    uint256 public constant MEMBERSHIP_PROOF = 0x06;
+    uint256 public constant NON_MEMBERSHIP_PROOF = 0x07;
```
aeryz, line 11: btw note that it might be a good idea to add `PACKET_TIMEOUTS` here as a future-proof for sui timeouts. Otherwise we might mistakenly override.

lib/proof-lens-light-client-types/src/client_state.rs
```diff
@@ -0,0 +1,124 @@
+use ibc_union_spec::ClientId;
+
+/// Representation of the client state of a proof lens client.
+///
+/// For a state lens client A->B->C, where the state lens is running on A and tracking C, the
```
aeryz: proof lense?

Cargo.toml
benluelo: they were duplicated

cosmwasm/core/src/tests.rs
benluelo: yep will remove, was just for debugging

cosmwasm/lightclient/proof-lens/src/raw_bytes.rs
benluelo: yeah ideally i want to have a separate codec for proofs, but that's too much lift for this pr, will restructure in a followup pr

PR #5422, opened 2026-02-01
fix(node): only use feemarket fee collector

benluelo:


uniond/app/ante.go
```diff
@@ -76,7 +76,8 @@ func NewAnteHandler(options HandlerOptions) (sdk.AnteHandler, error) {
 			options.FeeMarketBankKeeper,
 			options.FeegrantKeeper,
 			options.FeeMarketKeeper,
-			ante.NewDeductFeeDecorator(options.AccountKeeper, options.BankKeeper, options.FeegrantKeeper, options.TxFeeChecker),
+			nil,
+			// ante.NewDeductFeeDecorator(options.AccountKeeper, options.BankKeeper, options.FeegrantKeeper, options.TxFeeChecker),
```
PoisonPhang: Why remove the fallback fee decorator?
https://github.com/skip-mev/feemarket/blob/45c64e7781c8f154f610c11f4ad27944c63d8850/tests/app/ante.go#L45-L69

benluelo: because feemarket is always enabled, we don't need the fallback

PR #5427, opened 2026-02-02
Voyager otel logs

benluelo:


aeryz (dismissed):
Most of this is logging so i skimmed through only but ggs

PR #5461, opened 2026-05-15
fix(zkgm): wrong sender field in _timeoutCall

enlightened88:
_timeoutCall was passing call.contractAddress instead of call.sender to encodeCallCalldata. Both _executeCall and _acknowledgeCall use call.sender correctly only the timeout path had the wrong argument.

This causes the inner IBC module to receive a packet with an incorrect sender field on timeout, which can result in a wrong refund address or a stuck packet.

### DIFF
--- a/evm/contracts/apps/ucs/03-zkgm/Zkgm.sol
+++ b/evm/contracts/apps/ucs/03-zkgm/Zkgm.sol
@@ -862,7 +862,7 @@ contract UCS03Zkgm is
                 sourceChannelId: ibcPacket.sourceChannelId,
                 destinationChannelId: ibcPacket.destinationChannelId,
                 data: ZkgmLib.encodeCallCalldata(
-                    path, call.contractAddress, call.contractCalldata
+                    path, call.sender, call.contractCalldata
                 ),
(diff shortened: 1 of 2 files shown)


hussein-aitlahcen:
hi @enlightened88, very nice finding thanks a lot! would you mind writting a test companion for the fix? if not I'll proceed :).

enlightened88:
> hi @enlightened88, very nice finding thanks a lot! would you mind writting a test companion for the fix? if not I'll proceed :).

hi @hussein-aitlahcen  added the test companion in the latest commit!

hussein-aitlahcen:
> > hi @enlightened88, very nice finding thanks a lot! would you mind writting a test companion for the fix? if not I'll proceed :).
> 
> hi @hussein-aitlahcen added the test companion in the latest commit!

I don't see where the test is?

enlightened88:
> > > hi @enlightened88, very nice finding thanks a lot! would you mind writting a test companion for the fix? if not I'll proceed :).
> > 
> > 
> > hi @hussein-aitlahcen added the test companion in the latest commit!
> 
> I don't see where the test is?

done

hussein-aitlahcen (approved):

PR #5470, opened 2026-06-13
fix(unionvisor): resolve symlinks in bundle

benluelo:


PoisonPhang (approved):
tested with e2e, lgtm

PR #5471, opened 2026-06-14
fix(galoisd): correct signature/validator count check in Poll

### LINKED ISSUE #5428: Wrong if check condition,  error handling logic never trigger forever
https://github.com/unionlabs/union/blob/f15673cd2647721328f2fec5558afbe94cd391bc/galoisd/grpc/server.go#L308-L313

should change to
if len(req.TrustedCommit.Signatures) > len(req.TrustedCommit.Validators) 

and

if len(req.UntrustedCommit.Signatures) > len(req.UntrustedCommit.Validators)

bit2swaz:
fixes #5428

## Summary

both `Poll` implementations had two input-validation guards comparing each commit's `Signatures` slice length against itself, so the condition was always false and the "More signatures than validators" error never fired.
a malformed request with more signatures than validators would skip the check entirely and reach the prover. this fixes both comparisons to use `Validators` on the right-hand side. the same copy paste bug was in both server files; both are patched in this commit

## Changes

- `galoisd/grpc/server.go`: compare `Signatures` against `Validators` for both the trusted and untrusted commits
- `galoisd/grpc/bls12381_server.go`: same fix (the guard was copy pasted from `server.go` with the same typo)
- `galoisd/grpc/server_test.go`: regression test that `Poll` rejects a commit with more signatures than validators, covering both server types and both commit fields (4 sub-cases total)

## Testing

`go test ./grpc/...` in `galoisd\`. the new test fails before the fix (the dead guard is skipped so a zero-value server hits the job-slot check and returns `busy_building`), and passes after. also ran `go build ./grpc/...` and `go vet ./grpc/...`

### DIFF
--- a/galoisd/grpc/bls12381_server.go
+++ b/galoisd/grpc/bls12381_server.go
@@ -105,10 +105,10 @@ func (p *proverServerBls12381) Poll(ctx context.Context, pollReq *grpc.PollReque
 	if len(req.UntrustedCommit.Validators) > lightclient.MaxVal {
 		return nil, fmt.Errorf("The circuit can handle a maximum of %d validators", lightclient.MaxVal)
 	}
-	if len(req.TrustedCommit.Signatures) > len(req.TrustedCommit.Signatures) {
+	if len(req.TrustedCommit.Signatures) > len(req.TrustedCommit.Validators) {
 		return nil, fmt.Errorf("More signatures than validators")
 	}
-	if len(req.UntrustedCommit.Signatures) > len(req.UntrustedCommit.Signatures) {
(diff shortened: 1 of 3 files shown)


bit2swaz:
@benluelo any chances i can get a review on this PR?

benluelo:
you need to remove the merge commit and rebase your pr off of main instead

hussein-aitlahcen (approved):

bit2swaz:
rebased. i hope this works :)

benluelo (approved):

PoisonPhang (approved):

PR #5473, opened 2026-06-19
Fix galoisd signature count checks

### LINKED ISSUE #5428: Wrong if check condition,  error handling logic never trigger forever

samsamtrum:
Fixes #5428.

The Poll handlers compared each signatures length against itself, so the guard for more signatures than validators could never run. This updates both prover variants to compare signatures against validators.


kutluhaneth46 (commented):
## Review

This is a real validation bug, not a style nit:

```go
len(req.TrustedCommit.Signatures) > len(req.TrustedCommit.Signatures)  // always false
```

was a copy-paste tautology in both `server.go` and `bls12381_server.go`. Comparing signatures against **validators** is what the error string already claimed.

Clear correctness fix on the galoisd Poll path — LGTM.

PR #5474, opened 2026-06-22
fix(gno): update CanonicalPartSetHeader amino encoding to match gnolang/gno#5716

aronpark1007:
## Summary
Fixes the amino encoding of `CanonicalPartSetHeader` to match the field order and types defined in [gnolang/gno#5716](https://github.com/gnolang/gno/pull/5716).

Previously, `total` was encoded as `sint32` at field tag 2 and `hash` at tag 1, diverging from the upstream gno spec and causing signature verification failures against current gno testnet blocks.

## Changes

- **`lib/gno-types/src/part_set_header.rs`** — reorder fields in `CanonicalPartSetHeader`: `total` moved to tag 1 (`uint32`), `hash` to tag 2; update `canonicalize_part_set_header` accordingly
- **`lib/gno-types/src/vote.rs`** — update expected amino-encoded bytes in unit test to reflect corrected field ordering
- **`lib/gno-verifier/src/verify.rs`** — update testdata file references
- **`lib/gno-verifier/testdata`** — replace stale mainnet testdata with current test-13 block data (height 381822–381823)

## Test

Existing `verify_works` test in `lib/gno-verifier/src/verify.rs` runs against the new testnet13 testdata and passes with the corrected encoding.


benluelo (approved):
nice lgtm ty!

PR #5478, opened 2026-07-06
fix(gno): support inter-realm cross-calls in voyager tx and state plugins

aronpark1007:
## Summary

Rewrites gno tx plugin msg templates for the new inter-realm cross-call convention, fixes state-module parsing, and completes gno event-source parsing for `batch_send`/`packet_ack` which were previously unimplemented stubs.

### `voyager/plugins/transaction/gno/src/main.rs`

- Rewrite all datagram templates to `types.NewMsg*(...)` constructors under `cur realm` / `cross(cur)`
- Restore relayer wiring via `fee_recipient`
- Extract tx-failure classification into a `handle_gno_tx_failure()` helper (no behavior change — just deduping the existing `TxFailed` handling so it can be reused later; see test plan note below)

### `voyager/modules/state/gno/src/main.rs`

- Accept both old/new `parse_gno_string_object` return formats
- Find `PacketSend` event by type instead of by index
- Stop hardcoding the core realm path in `query_packet_by_hash`
- Search `PacketByHash` in bounded height windows instead of an unbounded query, configurable via `max_query_window`

### `voyager/plugins/event-source/gno/src/ibc_events.rs`

- Add `chunked_attr` helper to reassemble hex values gno splits across multiple attrs, used for `packet_data`, `acknowledgement`, `maker_msg`
- Implement `BatchSend { packet_hash, batch_hash, channel_id }` parsing (previously `{}` stub)
- Implement `PacketAck { packet_hash, source_channel_id, source_connection_id, source_connection_client_id, destination_channel_id, destination_channel_version, destination_connection_id, destination_connection_client_id, timeout_timestamp, acknowledgement, maker }` parsing (previously `{}` stub)
- Trim `PacketRecv` down to the fields gno actually emits (`packet_hash`, `destination_channel_id`, `maker_msg`)

### `voyager/plugins/event-source/gno/src/lib.rs`

- Wire up `BatchSend`/`PacketAck` construction (channel/connection state queries → `ibc_union_spec::event` assembly)
- Fix `PacketRecv` match arm to the trimmed field set

## Test plan

- `cargo +nightly-2025-12-05 check -p voyager-event-source-plugin-gno -p voyager-state-module-gno -p voyager-transaction-plugin-gno` passes cleanly (workspace-wide `cargo check` is blocked by an unrelated macOS/nightly `embed-commit` link-section issue, not touched by this branch)
- Added unit tests for the bounded-window math (`mk_windows`) covering exact multiples, remainders, and edge cases (window size 1, latest height 0/smaller than window)
- Manually cross-checked every parsed attribute key/type against the actual gno contract emit call sites (e.g. `emitPacketAckEvent`) to confirm they match
- Note: `handle_gno_tx_failure` doesn't yet handle `BroadcastTxCommitError::SimulationFailed` — that variant only exists once #5484 lands, tracked there


aronpark1007:
Right now the gno event interface defines every field the on-chain event emits, but most of them end up marked `_` (unused) at the point they're actually consumed. Looking at other chains, e.g. Sui, the event interface only defines the fields that are actually parsed/used, and the rest (like `packet` metadata) is filled in later via RPC queries. Would it make sense for gno to trim its event interfaces down to just the fields that are actually consumed, following the same pattern?

For example, `PacketAck`:
https://github.com/unionlabs/union/blob/75feaf26719058955842188f36d76965c2751acb/voyager/plugins/event-source/gno/src/ibc_events.rs#L84-L96

is only actually consumed like this (rest marked `_`):
https://github.com/unionlabs/union/blob/75feaf26719058955842188f36d76965c2751acb/voyager/plugins/event-source/gno/src/lib.rs#L1014-L1026

vs. Sui's equivalent, which only carries the fields it needs:
https://github.com/unionlabs/union/blob/75feaf26719058955842188f36d76965c2751acb/voyager/plugins/event-source/sui/src/events.rs#L119-L124

voyager/plugins/event-source/gno/src/ibc_events.rs
```diff
@@ -68,11 +68,21 @@ pub enum IbcEvent {
...
+    },
 
     PacketRecv {
         packet_hash: H256,
-        packet_data: Bytes,
+        destination_channel_id: ChannelId,
+        maker_msg: Bytes,
+    },
+
+    // TODO
```
benluelo: can remove todo

voyager/modules/state/gno/src/main.rs
```diff
@@ -798,11 +806,11 @@ impl StateModuleServer<IbcUnion> for Module {
...
 }
 
-// REVIEW: Do I need to do unescaping here?
 fn parse_gno_string_object(s: impl AsRef<str>) -> RpcResult<String> {
-    s.as_ref()
-        .strip_prefix("(\"")
-        .and_then(|s| s.strip_suffix("\" string)"))
-        .map(ToOwned::to_owned)
+    let s = s.as_ref();
+    // Handles both old format ("value" string) and new format
```
benluelo, line 1191: do we need to support both?

benluelo:
@aronpark1007 the event parsing can also be implemented more intelligently to use all of the emitted fields in the gno event, up to y'all how you want to emit them on-chain

voyager/plugins/event-source/gno/src/ibc_events.rs
aronpark1007: 86d2ba9525
Left on purpose earlier since the Gno-side event fields weren't finalized yet. They're confirmed now, so removed.

voyager/modules/state/gno/src/main.rs
aronpark1007: Yes, we need to support both — the two groups of gno functions have different return shapes:
- Query* functions (`QueryClientState`, `QueryConsensusState` ..) return a bare string, so vm/qeval formats them as ("value" string).
- Get* functions (e.g. `GetClientType`) return (T, error), so vm/qeval formats them as ("value" gno.land/p/.../types.T) followed by the error value.

The parser doesn't actually branch on which shape it's looking at — it only looks for the opening (" and the first closing "  to extract the value, so it already handles both without extra logic. Keeping as-is.

aronpark1007:
@benluelo 
Fixed the bptree proof verification issue — gno.land switched its main store from IAVL to a bptree store (gnolang/gno#5937), which uses a different ics23 spec (ics23:bptree vs IAVL), so our gno light client couldn't verify proofs anymore and relayers started failing with spec-mismatch errors.

Added a BPTREE_PROOF_SPEC/GNO_SPECS in lib/ics23 matching gno's actual bptree spec, wired it into the gno client-bootstrap so new clients use it, and along the way found+fixed an inverted max_depth check bug in ics23 that was silently broken until now. Added a regression test using a real proof pulled from topaz-1 (post-migration) to make sure it actually verifies.

tbruyelle:
For reference, here is what we had to do for our ts-relayer to support ics23:bptree spec https://github.com/allinbits/ibc-v2-ts-relayer/pull/24/changes

voyager/modules/state/gno/src/main.rs
benluelo: would probably be good to unify the query responses on the gno side.

lib/ics23/src/proof_specs.rs
```diff
@@ -68,6 +69,42 @@ pub const TENDERMINT_PROOF_SPEC: ProofSpec = ProofSpec {
...
+        max_prefix_length: result_unwrap!(PositiveI32AsUsize::new_const(1)),
+        empty_child: Bytes::new_static(&[
+            0xdb, 0xc1, 0xb4, 0xc9, 0x00, 0xff, 0xe4, 0x8d, 0x57, 0x5b, 0x5d, 0xa5, 0xc6, 0x38,
+            0x04, 0x01, 0x25, 0xf6, 0x5d, 0xb0, 0xfe, 0x3e, 0x24, 0x49, 0x4b, 0x76, 0xea, 0x98,
+            0x64, 0x57, 0xd9, 0x86,
+        ]),
+        hash: HashOp::Sha256,
+    },
+    max_depth: Some(result_unwrap!(
+        BoundedUsize::<1, { i32::MAX as usize }>::new_const(60)
```
benluelo, line 100: can use positivei32asusize here right?

benluelo (commented):
given the changes to the proof format, we'll either 1. need to implement a bptree verifier in solidity or 2. use proof lens clients for gno

aronpark1007:
@benluelo, Confirmed proof-lens bridging works correctly on-chain end to end. One issue we ran into along the way:

voyager-plugin-transaction-batch and voyager-plugin-transaction-batch-proof-lens can end up registered on the same chain, but the plain transaction-batch plugin can't handle events for a proof-lens client (encode_proof intentionally errors — "proofs for this client must be encoded for the underlying l1 client"). Since both plugins' interest_filter is effectively "take everything" (Any) or "take only these ids" (Many), leaving either on Any causes it to grab events meant for the other, causing a fatal error.

Working around this today by scoping both to Many with disjoint, explicitly-listed client_ids per chain — which works, but means manually maintaining an up-to-date id list any time a client is (re)created on that chain.

Is there a better way to handle this scoping, or would you be open to adding a Deny variant to ClientConfigsSerde (exclude a small set of ids, allow everything else) so we only need to track the proof-lens id(s) instead of every other client on the chain?

voyager/plugins/transaction-batch-proof-lens/src/lib.rs
```diff
@@ -190,7 +190,7 @@ impl ClientConfigs {
                     serde_json::to_string(&many.keys().map(|k| (k, ())).collect::<HashMap<_, _>>())
                         .unwrap();
 
-                format!("{clients_json} | has($client_id)")
+                format!("{clients_json} | has($client_id | tostring)")
```
aronpark1007, line 193: $client_id comes from the event as a JSON number, but object keys (and has()'s arg, for objects) are always strings — without tostring this errors instead of evaluating false, so Many was non-functional for any real (numeric) client_id.

If you think this should be fixed in a separate PR, happy to do that instead — let me know which you'd prefer.

lib/ics23/src/proof_specs.rs
aronpark1007: `max_depth`/`min_depth` on `ProofSpec` are `BoundedUsize`<1, ...> (min 1), not `PositiveI32AsUsize` (min 0) — so it doesn't quite drop in here. The min=1 bound is deliberate: proto's 0 means "unset" and maps to None, so Some(x) is reserved for x >= 1 to keep that mapping unambiguous.

Let me know if I'm misunderstanding something here!

voyager/plugins/transaction-batch-proof-lens/src/lib.rs
benluelo: leftover from IBC classic. would be nice to add a small test to make sure it works properly too

benluelo:
the normal batch plugin and the proof lens plugin are mutually exclusive; only one needs to be registered for a chain. having both of them registered doesn't make any sense

aronpark1007:
> the normal batch plugin and the proof lens plugin are mutually exclusive; only one needs to be registered for a chain. having both of them registered doesn't make any sense

You were right — we had transaction-batch and transaction-batch-proof-lens both registered on the same chain, which was unnecessary. Removed the plain transaction-batch entry for that chain and reverted the proof-lens one back to Any. Confirmed bidirectional transfer works end-to-end with this simpler setup.

Thank you for the quick feedback!

voyager/plugins/transaction-batch-proof-lens/src/lib.rs
aronpark1007: To actually prove this works (not just that the string looks right), I think we'd need to add jaq-core/jaq-json/jaq-std as dev-dependencies so the test can compile+run the generated filter against real numeric client_ids and assert on the result — a string-content check (e.g. contains("tostring")) wouldn't catch this class of bug, since the original issue was in filter execution behavior, not something visible from the source string.

These aren't new to the workspace — voyager-core already depends on the same versions for the real interest-filter engine — so this would just be reusing an already-vetted crate, scoped to dev-dependencies only (test-only, not shipped in the plugin binaries).

Are you okay with adding that, or would you rather we go with a lighter-weight check instead?

aronpark1007:
## gno tx plugin error handling

Previously *every* gno tx failure was retried forever. This PR classifies known `errors.gno` failures into drop-vs-retry:

| Error string | Handling | Reason |
|---|---|---|
| packet already acknowledged | Dropped | already finalized elsewhere |
| acknowledgement already written | Dropped | already finalized elsewhere |
| packet commitment not found | Dropped | already acked/timed out |
| packet already received | Dropped | duplicate recv |
| packet timeout expired | Dropped | dest-side timeout already passed |
| packet commitment already exists | Dropped | duplicate send |
| batch packets not found | Dropped | batch already processed |
| batch receipts not found | Dropped | batch already processed |
| port already registered | Dropped | idempotent, already done |
| caller not authorized to write ack / ...send packet | Dropped (error-logged) | signer/config bug |
| rlm does not match crossing frame / rlm is not proxy realm | Dropped (error-logged) | wiring bug |
| not enough packets / batch must use the same channel | Dropped (error-logged) | msg construction bug |
| invalid counterparty connection id / ...channel id | Dropped (error-logged) | msg construction bug |
| timeout must be set | Dropped (error-logged) | malformed packet |
| acknowledgement mismatch / ...count mismatch / ...cannot be empty | Dropped (error-logged) | malformed ack |
| client not found / connection not found / channel not found / port not found | Dropped (error-logged) | stale/wrong id (ids are never deleted) |
| invalid connection state / invalid channel state | **Retry (unconfirmed)** | might be mid-handshake race, might be stale msg |
| client is not active | **Retry (unconfirmed)** | frozen/expired client, unsure if recoverable |
| packet receipt not found | **Retry (unconfirmed)** | might be an indexing race |
| client state / consensus state / membership proof not found` | **Retry (unconfirmed)** | might be an indexing race |
| packet timeout not reached | Retry (intentional) | should self-resolve once time passes |
| packet combined data too large | Retry (out of scope) | should split batch instead, follow-up |

@benluelo, Would appreciate confirmation on the **Retry (unconfirmed)** rows — can any of these actually be permanent, or are they always transient races?

benluelo:
looks fine to me, any issues will be sorted out as Voyager is actually run anyways.

voyager/plugins/transaction-batch-proof-lens/src/lib.rs
benluelo: should be possible to just build an interest filter and run this? https://github.com/unionlabs/union/blob/773eb5d3e7682680539a49986a8e63ffb00898eb/lib/voyager-core/src/filter.rs#L129

aronpark1007: 0d846e5b58

voyager/modules/state/gno/src/main.rs
aronpark1007: This isn't something we can drop even after gno-ibc's functions all move to (value, error), since qeval's output format itself hasn't changed.

moul:
fyi, https://github.com/gnolang/gno/pull/5937 is merged.

PR #5484, opened 2026-08-24
feat(gno): support simulated and dynamic gas pricing for tx fees

aronpark1007:
## Summary
- Adds two new gas-pricing modes for the gno transaction plugin alongside the existing fixed fee: `simulate` (dry-run the tx via `.app/simulate` to get the actual gas used, then price it with the fixed linear formula) and `dynamic` (query gno's own protocol-level `auth/gasprice`, an EIP-1559-style base fee recomputed per block from actual usage).
- `Config.gas_config` selects the mode (`fixed` / `simulate` / `dynamic`); the legacy `fee` field is still accepted for backward compatibility (deprecated, mapped to `fixed`).
- Denom resolution is fail-fast: if the chain reports no gas-price denom and no `denom` override is configured, fee construction now errors instead of silently producing a denom-less coin string.

## Test plan
- `cargo test -p gno-client`: parser/pricing/denom-fallback unit tests, including regression coverage for pricing off the post-multiplier gas limit.
- `cargo test -p voyager-transaction-plugin-gno`: config deserialization (new `gas_config` schema, legacy `fee` fallback, both-set precedence, missing-both error).
- Verified on union-devnet after migrating the gno light client contract — confirmed working on-chain.


aronpark1007:
TODO once #5478 lands: add a `BroadcastTxCommitError::SimulationFailed` arm in the gno tx plugin so it goes through the same permanent-failure handling as `TxFailed`. Right now a simulation failure from `simulate`/`dynamic_gas_price` would just retry forever.

